PR #5, opened 2018-08-03
Parse lockfile recursively

miiila:
- 💥 Tests written and linted [ℹ︎]() 

### What this does

Introduces a recursive algorithm for creation of dependency tree from `package-lock.json` and `package.json`.

### Notes for the reviewer
I did my best to comment the algorithm, but feel free to suggest any naming or other structural improvements.

### More information

- [SC-5885]()

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -1,0 +1,10 @@
+#!/usr/bin/env node
+const inspect = require('../dist/index')
+
+inspect.buildDepTreeFromFiles('./', 'package.json', 'package-lock.json')
+  .then((tree) => {
+    console.log(JSON.stringify(tree));
+  })
+  .catch((e) => {
+    console.log(e);
+  });
--- a/bin/index.js
+++ b/bin/index.js
@@ -5,6 +5,4 @@ inspect.buildDepTreeFromFiles('./', 'package.json', 'package-lock.json')
   .then((tree) => {
     console.log(JSON.stringify(tree));
   })
-  .catch((e) => {
-    console.log(e);
-  });
+  .catch(console.error);
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,7 +1,7 @@
 import 'source-map-support/register';
-import * as fs from "fs";
-import * as path from "path";
-import * as _ from "lodash";
+import * as fs from 'fs';
+import * as path from 'path';
+import * as _ from 'lodash';
 
 export default function parseLockFile(root, targetFilePath, lockFilePath, options) {
   if (!root || !lockFilePath || !lockFilePath) {
@@ -54,7 +54,7 @@ async function buildDepTree(targetFileRaw, lockFileRaw, options) {
 
   async function buildSubTreeRecursive(dep: string, depKeys: string[]) {
 
-    const depTree = {
+    const depSubTree = {
       dependencies: {},
       name: dep,
       version: undefined,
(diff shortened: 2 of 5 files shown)


lib/index.ts
```diff
@@ -40,46 +42,63 @@ function buildDepTree(targetFileRaw, lockFileRaw, options) {
...
+      // repeat the process for dependencies of looked-up dep
+      if (deps[dep].requires) {
+        Object.keys(deps[dep].requires).forEach(async (subDep) => {
+          depSubTree.dependencies[subDep] = await buildSubTreeRecursive(subDep, [...depKeys, subDep]);
+
+        });
+        return depSubTree;
+      } else {
+        // no more deps, return tree
+        return depSubTree;
```
lili2311: don't think you need an `else` here since you return in the statement above

lib/index.ts
```diff
@@ -40,46 +42,63 @@ function buildDepTree(targetFileRaw, lockFileRaw, options) {
...
+      }
+    } else {
+      // tree was walked to the root and dependency was not found
+      if (!depKeys.length) {
+        throw new Error(`Dependency ${dep} was not found in package-lock.json.`);
```
lili2311: Can we make this message more verbose and let the user know that it looks like their package.json and their lockfile are out of sync and give them some advice how to sync them up?

lib/index.ts
```diff
@@ -40,46 +42,63 @@ function buildDepTree(targetFileRaw, lockFileRaw, options) {
...
+
+    // If exists and looked-up dep is there
+    if (deps && deps[dep]) {
+      // update the tree
+      depSubTree.version = deps[dep].version;
+      // repeat the process for dependencies of looked-up dep
+      if (deps[dep].requires) {
+        Object.keys(deps[dep].requires).forEach(async (subDep) => {
+          depSubTree.dependencies[subDep] = await buildSubTreeRecursive(subDep, [...depKeys, subDep]);
+
```
lili2311: Can you run the linter on this, looks like they might be some linting fixes to add.

lib/index.ts
miiila: You are right, I'll move return to the separate statement.

lib/index.ts
miiila: Will try :-)

lib/index.ts
miiila: `tslint` doesn't complain.

package.json
```diff
@@ -25,11 +25,12 @@
   "devDependencies": {
     "@types/node": "10.5.5",
     "@types/sinon": "5.0.1",
+    "semantic-release": "^8.2.0",
```
adrukh: Please remove :)

lib/index.ts
lili2311: Hah strange!

miiila: Didn't complain :-D It does now.

bin/index.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+const inspect = require('../dist/index')
+
+inspect.buildDepTreeFromFiles('./', 'package.json', 'package-lock.json')
+  .then((tree) => {
+    console.log(JSON.stringify(tree));
+  })
+  .catch((e) => {
```
darscan: `.catch(console.error);` might be better here

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "snyk-nodejs-lockfile-parser",
   "description": "Generate a dep tree given a lockfile",
   "main": "dist/lib/index.js",
+  "bin": {
+    "parse": "./bin/index.js"
```
darscan: I think `"parse"` is probably too generic for a bin that will be installed globally

lib/index.ts
```diff
@@ -96,3 +73,25 @@ function getDepPath(depKeys: string[]) {
 
   return depPath;
 }
+
+export function buildDepTreeFromFiles(root, targetFilePath, lockFilePath, options) {
+  if (!root || !lockFilePath || !lockFilePath) {
+    throw new Error('Missing required parameters for parseLockFile()');
+  }
+  // TODO: validate only valid options were passed in
```
darscan: I might drop this TODO.. just noise

package.json
darscan: Not sure what a better name might be.. maybe "parse-npm-lockfile"??

PR #7, opened 2018-08-03
Feat/enable all node

lili2311:
### What this does
Add node 4 & 6 to be tested


adrukh:
Wow this was easy!
Care to enable 4 and 6 on appveyor as well?

lili2311:
How do I do that?

PR #13, opened 2018-08-12
Chore/more types

michael-go:
This PR mostly adds more type annotations and fixes `tslint.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "test": "npm run lint && npm run unit-test",
     "unit-test": "tap test/lib -R=spec --timeout=300 --node-path ts-node --test-file-pattern '/\\.[tj]s$/'",
-    "lint": "tslint lib/**/*.ts",
+    "lint": "tslint -p tsconfig.json",
     "build": "tsc",
     "build-watch": "tsc -w",
     "prepare": "npm run build",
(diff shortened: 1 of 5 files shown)


michael-go:
I think it would be better to rename `targetFile` to `manifestFile` or even `packageJson` - WDYT?

package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "test": "npm run lint && npm run unit-test",
     "unit-test": "tap test/lib -R=spec --timeout=300 --node-path ts-node --test-file-pattern '/\\.[tj]s$/'",
-    "lint": "tslint lib/**/*.ts",
+    "lint": "tslint -p tsconfig.json",
```
miiila: Just a quick question - will it still be linting `.ts` files only? (https://github.com/snyk/nodejs-lockfile-parser/blob/4e08bcbdf8836229e03841cb5ae1d9cd2cc9a739/tsconfig.json#L14)

I tried to lookup this in the docs, but haven't found it.

tslint.json
```diff
@@ -4,31 +4,14 @@
...
-          {
-            "SwitchCase": 1
-          }
-        ],
-        "max-depth": [1, 3],
-        "max-len": [1, 80],
-        "max-statements": [1, 15],
-        "no-use-before-define": [2, "nofunc"],
-        "object-curly-spacing": [2, "never"],
+        "trailing-comma": [true, {"multiline": "always"}],
```
miiila: I'd just want to understand - is this a set of rules we are using in other TypeScript libraries?

package.json
michael-go: by default it will lint also .js, need to add something like `-e '**/*.js'` to skip `.js` - but not needed here

tslint.json
michael-go: it depends - we still don't have a standard for this. In the last TypeScript project @gjvis decided to leave it almost blank with:
```json
{
    "defaultSeverity": "error",
    "extends": [
        "tslint:recommended"
    ],
    "jsRules": {},
    "rules": {
        "quotemark": {
            "options": "single"
        }
    },
    "rulesDirectory": []
}
```

in this PR, I tried to leave the original intension

PR #15, opened 2018-08-13
feat: add support for different type of dependencies

miiila:

### What this does

Add support for dev and prod types of dependencies.

### Notes for the reviewer

Unfortunately tests are failing because of the way fixtures were generated. Will be discussed tomorrow with @lili2311 and find a proper solution.

https://github.com/snyk/nodejs-lockfile-parser/pull/15/commits/6a045fc07c25cc7520cf007d61fbe922d8592533 may sound weird, but it's a workaround for a bug found in https://github.com/snyk/resolve-deps/, which causes only top level dev-dependency being marked as `dev`, but all children remaining `prod`.

### More information

- [ SC-5960](https://snyksec.atlassian.net/browse/SC-5960)


test/lib/fixtures/goof/dep-tree_small.json
```diff
@@ -1,84 +1,101 @@
 {
   "name": "goof",
   "version": "0.0.3",
```
lili2311: Can we keep the file names concistent with - instead of _? Also please can we name the trees to match the test cases? Rather than small and just dep-tree an we calle them something like:
dep-tree-no-dev-deps and dep-tree-with-dev-deps etc so it is very clear which test each one matches

miiila: Good point, by small, I tried to suggest it's a subset of whole depTree generate by resolve-deps, but it's probably not needed anymore.

lib/index.ts
```diff
@@ -3,20 +3,29 @@ import * as fs from 'fs';
...
+  dev = 'dev',
+}
 
 interface PkgTree {
   name: string;
   version: string;
   dependencies?: {
     [dep: string]: PkgTree;
   };
+  depType?: DepType;
```
michael-go: `depType` is not currently used by other modules .. we can leave it for feature-parity with `snyk-resolve-deps`.. but don't have to

lib/index.ts
```diff
@@ -40,12 +50,21 @@ async function buildDepTree(targetFileRaw: string, lockFileRaw: string): Promise
     depTree.dependencies[dep] = await buildSubTreeRecursive(dep, [], lockFile);
   }));
 
+  if (includeDev && targetFile.devDependencies) {
+    const topLevelDevDeps = Object.keys(targetFile.devDependencies);
+    depTree.hasDevDependencies = true;
+    await Promise.all(topLevelDevDeps.map(async (dep) => {
```
michael-go: relevant also for the prod dependencies - why do we need the `Promise.all()` and `await` here? it will all still run on a single thread and doesn't do any async IO. So it only creates unnecessary overhead I think

lib/index.ts
```diff
@@ -30,6 +39,7 @@ async function buildDepTree(targetFileRaw: string, lockFileRaw: string): Promise
 
   const depTree: PkgTree = {
     dependencies: {},
+    hasDevDependencies: false,
```
lili2311: Please use targetFile.devDependencies as this seems to just be metadata about the package.json

lib/index.ts
miiila: I believe we shouldn't look at this library through the optic of other applications. It should make sense as a standalone module, therefore I'd keep an info whether dependency is `prod` or `dev`.

We can definitely discuss the right naming or structure (e.g. should we follow `package-lock.json` and use `"dev": true,`?), but I'd like to keep an info there.

lib/index.ts
miiila: I was following @darscan suggestion `do it async from the beginning, because it's much harder to do it later` and I agree with that - overhead now can save us a toil in the future, which is too far away to say, whether I/O operations will be introduced or not.

test/lib/index.ts
```diff
@@ -22,6 +22,31 @@ test('Parse npm package-lock.json', async (t) => {
   t.deepEqual(depTree, JSON.parse(expectedDepTree), 'Tree generated as expected');
 });
 
+test('Parse npm package-lock.json with devDependencies', async (t) => {
```
lili2311: `hasDevDependencies` can we assert here it is true?

miiila: @lili2311 It's in a fixture - https://github.com/snyk/nodejs-lockfile-parser/pull/15/files/355f3fb209d6e24d9b36929c328db87e5a7203cf#diff-00d698a5881a0baea201915c8519647cR4

lib/index.ts
lili2311: Please lets keep it is it handy for debugging where each dep came from even if it is not yet used anywhere, I see this being useful to show in registry actually against each vuln/dep and would like to suggest we add it there soon

PR #18, opened 2018-08-22
Add support for yarn

miiila:

### What this does

Adds support for `yarn.lock`

### Notes for the reviewer

Because of the differencies in `yarn` and `npm` resolution of dependecies, tree has to be different.

In order to have better organisation of code, Typescript classes were used to encapsulate behaviour in corresponding units and to extract shared behaviour into a parent class. In a same time, I haven't introduced any state in order to stick to pure functions without side effects.

There's one exception and it's caused by lack of Node v4 support for yarn.lock parser. In that case, state is used to persist required package (https://github.com/snyk/nodejs-lockfile-parser/pull/18/files#diff-5259b645479c0f7e10680520bf8e7ecbR33).

You might wonder about two exceptions being thrown for similar root cause. I thought about it quite a while and realised, they have a different meaning.

https://github.com/snyk/nodejs-lockfile-parser/pull/18/files#diff-5259b645479c0f7e10680520bf8e7ecbR34 is meant to be for further developers of the library and it should save them nasty surprise, while https://github.com/snyk/nodejs-lockfile-parser/pull/18/files#diff-13b5b151431c7e7a17f273559ed212d5R40 is for users of whole library. Since this behaviour is expected, it shouldn't (IMO) wait for Error thrown by constructor.

### More information

- [SC-5891](https://snyksec.atlassian.net/browse/SC-5891)
- [Link to documentation]()


lib/index.ts
```diff
@@ -140,8 +192,68 @@ async function buildSubTreeRecursive(
...
+  }
+
+  return depSubTree;
+}
+
+function parseLockFile(lockFileRaw: string, lockfileType: LockfileType, includeDev: boolean): PackageLock | YarnLock {
+  let lockfile: PackageLock | YarnLock ;
+  switch (lockfileType) {
+    case LockfileType.npm:
+      lockfile = JSON.parse(lockFileRaw);
```
lili2311: What if the parsing fails? Do we have a test with an invalid json file? Would like to see us handle this with try/catch

lib/index.ts
```diff
@@ -140,8 +192,68 @@ async function buildSubTreeRecursive(
...
+      if (!(lockfile as PackageLock).dependencies && !includeDev) {
+        throw new Error("No 'dependencies' property in package-lock.json");
```
lili2311: This error message is `package-lock.json` specific, now we handle both filetypes let's actually mention which one we are talking about.

test/lib/index.ts
```diff
@@ -146,3 +146,15 @@ test('Performance: Parse big npm package-lock.json with cyclic deps and dev-deps
   );
   t.deepEqual(depTree.name, 'trucolor', 'Tree is created correctly');
 });
+
+test('Parse yarn yarn.lock', async (t) => {
```
lili2311: I would like to see a test where we handle invalid files to parse as JSON for both of the lockfile types.

lib/index.ts
```diff
@@ -150,6 +262,9 @@ async function buildDepTreeFromFiles(
     throw new Error('Missing required parameters for buildDepTreeFromFiles()');
   }
 
+  const lockFileType = lockFilePath.endsWith('package-lock.json') ?
```
lili2311: This is not specific enough, we have to specifically check for yarn.lock. What if we get passed some other random file name?

lib/index.ts
```diff
@@ -163,5 +278,5 @@ async function buildDepTreeFromFiles(
   const manifestFileContents = fs.readFileSync(manifestFileFullPath, 'utf-8');
```
lili2311: What if this fails? Can it fail? We don't seem to try/catch much around here.

lili2311:
https://github.com/snyk/nodejs-lockfile-parser/blob/20b1b7f8ef99e167c1e5a93f549149209ee11a88/lib/index.ts#L261 a little mistake here, we want to check for `targetFilePath` instead of twice checking the `lockFilePath`

lib/index.ts
```diff
@@ -54,12 +84,13 @@ export {
   buildDepTreeFromFiles,
   PkgTree,
   DepType,
+  LockfileType,
 };
 
 async function buildDepTree(
```
lili2311: We should be determining the lockfile type in the `buildDepTree` function, given just the target and lockfile and some options. We shouldn't expect the lockfile type to be derived before calling our lib given we can easily determine this.

lib/index.ts
```diff
@@ -140,8 +192,68 @@ async function buildSubTreeRecursive(
...
+  };
+
+  const depKey = `${searchedDep.name}@${searchedDep.version}`;
+
+  const dep = _.get(lockFile.object, depKey);
+
+  if (!dep) {
+    throw new Error(`Dependency ${depKey} was not found in yarn.lock.
+      Your package.json and yarn.lock are probably out of sync.
+      Please run npm install and try to parse the log again.`);
```
lili2311: This sentence doesn't quite make sense to me: `Please run npm install and try to parse the log again.`

lib/index.ts
miiila: I have another point of view - every caller of our library will already know, if they want to pass `yarn.lock` or `package-lock.json` (because they already had to read file, since  `buildDepTree` requires `string`. Therefore, any effort to derive type of a lockfile would be unnecessary and can add some performance overhead.

That's the reason why I added this functionality to `buildDepTreeFromFiles`, because it can be called with any path.

lib/index.ts
miiila: Because it's in `package-lock` branch, `yarn` specific error is few lines below (https://github.com/snyk/nodejs-lockfile-parser/pull/18/files#diff-13b5b151431c7e7a17f273559ed212d5R252)

lib/index.ts
miiila: In progress :-)

test/lib/index.ts
miiila: Tests added

lib/index.ts
miiila: It shouldn't™️ fail, because we are verifying both path via `fs.existsSync` few lines above. If it would fail despite that, it would just throw an error from the function.

lib/index.ts
miiila: typo 😇

lib/index.ts
```diff
@@ -259,12 +267,19 @@ function parseLockFile(lockFileRaw: string, lockfileType: LockfileType, includeD
...
-  const lockFileType = lockFilePath.endsWith('package-lock.json') ?
-    LockfileType.npm : LockfileType.yarn;
+  let lockFileType: LockfileType;
+  if (lockFilePath.endsWith('package-lock.json')) {
+    lockFileType = LockfileType.npm;
+  } else if (lockFilePath.endsWith('yarn.lock')) {
+    lockFileType = LockfileType.yarn;
+  } else {
+    throw new Error(`Unknown lockfile ${lockFilePath}.
+      Please provide either package-lock.json or yarn.lock.`);
```
lili2311: ❤️

lili2311:
`lib/parsers/parser.ts` should probably be `lib/parsers/index.ts` to follow our normal standards and stop imports like `parsers/parser`

lib/index.ts
lili2311: I did not mean a caller within the library but me calling this lib from Registry, maybe I misunderstand but I the way I will be calling this lib from registry is 
` const depTree = await lockFileParser.buildDepTree(rawManifetFile.text, rawLockFile); ` which is the call signature we all agreed really early on.

lib/index.ts
```diff
@@ -1,167 +1,90 @@
...
-  manifestFileContents: string, lockFileContents: string, includeDev = false): Promise<PkgTree> {
-
-  const lockFile: Lockfile = JSON.parse(lockFileContents);
-  const manifestFile: ManifestFile = JSON.parse(manifestFileContents);
+  manifestFileContents: string, lockFileContents: string,
+  includeDev = false, lockfileType?: LockfileType): Promise<PkgTree> {
 
-  if (!manifestFile.dependencies && !includeDev) {
-    throw new Error("No 'dependencies' property in package.json");
+  if (!lockfileType) {
```
lili2311: I would prefer we don't assume a type but throw, unexpected errors during parsing a wring file type will be worse than use being extra strict and provide back a very clear useful message.

lib/index.ts
miiila: Hmm, IIRC, initial signature was `buildDepTree(rawManifestFile, rawLockFile, options`, which was moved to `buildDepTree(rawManifestFile, rawLockFile, includeDev = false` and I now added `buildDepTree(rawManifestFile, rawLockFile, includeDev = false, lockfileType = 'npm')`*.

My point is, that even in registry, I assume you'd already know, which file you wanna parse (since we are looking specifically for `package-lock.json` or `yarn.lock`). Therefore, adding a functionality to determine a type of a lockfile is superfluous. Or am I missing something?

\* implemented via if statement

PR #20, opened 2018-09-02
perf: faster package-lock tree creation

miiila:
🚧 **DO NOT MERGE** 🚧

### What this does

Big `package-lock.json` parsing was slow. This PR introduces better parsing algorithm, which speeds up parsing a lot. 

Proper testing is still needed to be sure that functionality is the same. Tests are passing*, but they can't cover everything. However, it's pretty hard to verify what's right for the output with ~million lines.

\* I am aware about one failing, it's a topic for the discussion about the meaning of `depType` property.

*Before:*
```
$ time parse-nodejs-lockfile > /dev/null
parse-nodejs-lockfile > /dev/null  30.25s user 1.24s system 123% cpu 25.522 total
```
*After:*
```
$ time node ../lockfile-parser/bin > /dev/null
node ../lockfile-parser/bin > /dev/null  0.39s user 0.06s system 103% cpu 0.436 total
```

*30.25s* vs  *0.39s*

### DIFF
--- a/lib/parsers/package-lock-parser.ts
+++ b/lib/parsers/package-lock-parser.ts
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import * as graphlib from 'graphlib';
 import * as uuid from 'uuid/v4';
+import {setImmediatePromise} from '../set-immediate-promise';
 
 import {LockfileParser, PkgTree, Dep, DepType, ManifestFile,
   getTopLevelDeps, Lockfile, LockfileType} from './';
@@ -114,7 +115,7 @@ export class PackageLockParser implements LockfileParser {
         // if the top level dependency is dev, all children are dev
         depTree.dependencies[dep.name] = dep.dev ?
           this.setDevDepRec(_.cloneDeep(depTrees[depName])) : depTrees[depName];
-        await this.setImmediatePromise();
+        await setImmediatePromise();
       }
     }
     return depTree;
@@ -147,40 +148,40 @@ export class PackageLockParser implements LockfileParser {
   private removeCycle(
     cycle: string[], depMap: DepMap, depGraph: graphlib.Graph): CycleStartMap {
 
+    /* FUNCTION DEFINITION
+    To keep an order of algorithm steps readable, function is defined on-the-fly
+    Arrow function is used for calling `this` without .bind(this) in the end
+    */
+    const acyclicDuplicationRec = (node, traversed, currentCycle, nodeCopy) => {
+      // 2. For every cyclic dependency of entry node...
+      const edgesToProcess = (depGraph.inEdges(node) as graphlib.Edge[])
+      .filter((e) => _.includes(currentCycle, e.v));
+      for (const edge of edgesToProcess) {
+        // ... create a duplicate of the dependency...
+        const child = edge.v;
+        const dependencyCopy = this.cloneNodeWithoutEdges(child, depMap, depGraph);
+        // ...and connect it with the duplicated entry node
+        depGraph.setEdge(dependencyCopy, nodeCopy);
+        // 3.a If edge goes to already-visited dependency, end of cycle is found;
+        if (_.includes(traversed, child)) {
+          // update metadata and do not continue traversing
+          depMap[dependencyCopy].cyclic = true;
+        } else {
+          // 3.b Follow the edge and repeat the process, storing visited dependency-paths
+          acyclicDuplicationRec(child, [...traversed, node], currentCycle, dependencyCopy);
+          // All non-cyclic dependencies of duplicated node need to be updated.
+          this.cloneAcyclicNodeEdges(child, dependencyCopy, cycle, depGraph,
+          {inEdges: true, outEdges: false});
+        }
+      }
+    };
+
     const cycleStarts: CycleStartMap = {};
     // For every node in a cycle:
     for (const start of cycle) {
       // 1. Create a uniqe duplicate of entry node (without edges)
       const clonedNode = this.cloneNodeWithoutEdges(start, depMap, depGraph);
       cycleStarts[start] = clonedNode;
-
-      /* FUNCTION DEFINITION
-        To keep an order of algorithm steps readable, function is defined on-the-fly
-        Arrow function is used for calling `this` without .bind(this) in the end
-      */
-      const acyclicDuplicationRec = (node, traversed, currentCycle, nodeCopy) => {
-        // 2. For every cyclic dependency of entry node...
-        const edgesToProcess = (depGraph.inEdges(node) as graphlib.Edge[])
-          .filter((e) => _.includes(currentCycle, e.v));
-        for (const edge of edgesToProcess) {
-          // ... create a duplicate of the dependency...
-          const child = edge.v;
-          const dependencyCopy = this.cloneNodeWithoutEdges(child, depMap, depGraph);
-          // ...and connect it with the duplicated entry node
-          depGraph.setEdge(dependencyCopy, nodeCopy);
-          // 3.a If edge goes to already-visited dependency, end of cycle is found;
-          if (_.includes(traversed, child)) {
-            // update metadata and do not continue traversing
-            depMap[dependencyCopy].cyclic = true;
-          } else {
-            // 3.b Follow the edge and repeat the process, storing visited dependency-paths
-            acyclicDuplicationRec(child, [...traversed, node], currentCycle, dependencyCopy);
-            // All non-cyclic dependencies of duplicated node need to be updated.
-            this.cloneAcyclicNodeEdges(child, dependencyCopy, cycle, depGraph,
-              {inEdges: true, outEdges: false});
-          }
-        }
-      };
 
       // CALL of previously defined function
       acyclicDuplicationRec(start, [], cycle, clonedNode);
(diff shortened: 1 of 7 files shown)


lib/parsers/package-lock-parser.ts
```diff
@@ -41,66 +51,154 @@ export class PackageLockParser implements LockfileParser {
...
 
-    const topLevelDeps: Dep[] = getTopLevelDeps(manifestFile, includeDev);
-
     // asked to process empty deps
     if (_.isEmpty(manifestFile.dependencies) && !includeDev) {
       return depTree;
     }
 
+    // prepare a flat map, where keys are path, where dependencies are listed
+    const depMap: DepMap = await this.flattenLockfile(lockfile);
```
michael-go: pretty sure all this `async/await` is not needed. given no async IO is happening in this loop, the async/await would still block the even-loop until done

miiila: Yep, that's one of the thing I need your help and expertise - it's not needed now, but this package can end up being used in a deployed service, so we need to be sure we won't block event loop there. I know (from your talk) can push event loop to tick by `setImmediate()`, but I'd like to discuss the right approach.

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,234 @@ export class PackageLockParser implements LockfileParser {
...
+      //   updated to be connected with the duplicated one
+
+      // dependants
+      const edges = depGraph.nodeEdges(start) as graphlib.Edge[]; // node has to have edges
+      const entryEdges = edges.filter((e) => !_.includes(cycle, e.w));
+      for (const edge of entryEdges) {
+        depGraph.setEdge(newNode, edge.w);
+      }
+      // dependencies
+      const outEdges = edges.filter((e) => !_.includes(cycle, e.v));
```
Kirill89: `!_.includes(cycle, e.v)` -> `!cycle.includes(e.v)`?

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,234 @@ export class PackageLockParser implements LockfileParser {
...
+
+    return cycleStarts;
+  }
+
+  private createGraphOfDependencies(depMap: DepMap): graphlib.Graph {
+    const depGraph = new graphlib.Graph();
+    for (const depName of Object.keys(depMap)) {
+      depGraph.setNode(depName, depName);
+    }
+    for (const [depPath, dep] of _.entries(depMap)) {
```
Kirill89: `_.entries(` -> `Object.entries(`?
If you don't mind, I prefere to use native JS features whenever it's possible instead of lodash.

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,234 @@ export class PackageLockParser implements LockfileParser {
...
+      for (const subDepPath of depGraph.predecessors(depKey)) {
+        const subDep = depTrees[subDepPath];
+        dep.dependencies[subDep.name] = subDep;
       }
-      // dependency was not found on a current path, remove last key (move closer to the root) and try again
-      // visitedDepPaths can be passed by a reference, because traversing up doesn't update it
-      return this.buildSubTreeRecursiveFromPackageLock(depName, lockfilePath.slice(0, -1), lockFile, depPath, isDev);
+      delete dep.requires;
+      depTrees[depKey] = {...dep as PkgTree};
+      await setImmediatePromise();
```
Kirill89: Can you please add comment to this line? It is not so obvious why we need it.

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,240 @@ export class PackageLockParser implements LockfileParser {
...
     // asked to process empty deps
     if (_.isEmpty(manifestFile.dependencies) && !includeDev) {
       return depTree;
     }
 
-    await Promise.all(topLevelDeps.map(async (dep) => {
-      depTree.dependencies[dep.name] = await this.buildSubTreeRecursiveFromPackageLock(
-        dep.name, ['dependencies'], packageLock, [], dep.dev);
-    }));
+    // prepare a flat map, where dependency path is a key to dependency object
```
michael-go: comment about the `if` above, unrelated to this PR: what will happen if both `dependencies` and `devDepenedencies` are empty?

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,240 @@ export class PackageLockParser implements LockfileParser {
...
+    for (const dep of topLevelDeps) {
+      // if any of top level dependencies is a part of cycle
+      // it now has a different item in the map
+      const depName = cycleStarts[dep.name] ? cycleStarts[dep.name] : dep.name;
+      if (depTrees[depName]) {
+        // if the top level dependency is dev, all children are dev
+        depTree.dependencies[dep.name] = dep.dev ?
+          this.setDevDepRec(_.cloneDeep(depTrees[depName])) : depTrees[depName];
+      }
+    }
```
michael-go: would add the `seImmediatePromise()` call in this loop too, as the `setDevDepRec` might be blocking (maybe call it only if `dep.dev`)

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,240 @@ export class PackageLockParser implements LockfileParser {
...
+
+    // transform depMap to a map of PkgTrees
+    const depTrees: {[depPath: string]: PkgTree} = await this.createDepTrees(depMap, depGraph);
+
+    // get trees for dependencies from manifest file
+    const topLevelDeps: Dep[] = getTopLevelDeps(manifestFile, includeDev);
+    for (const dep of topLevelDeps) {
+      // if any of top level dependencies is a part of cycle
+      // it now has a different item in the map
+      const depName = cycleStarts[dep.name] ? cycleStarts[dep.name] : dep.name;
```
michael-go: can be simplified to: 
```javascript
const depName = cycleStarts[dep.name] || dep.name;
```

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,240 @@ export class PackageLockParser implements LockfileParser {
...
-      if (!lockfilePath.length) {
-        throw new Error(`Dependency ${depName} was not found in package-lock.json.
-          Your package.json and package-lock.json are probably out of sync.
-          Please run "npm install" and try again.`);
+      depPath.pop();
+    }
+
+    if (!depMap[depName]) {
+      throw new Error(`Dependency ${depName} was not found in package-lock.json.
+        Your package.json and package-lock.json are probably out of sync.
```
michael-go: when playing with some packages, this error happened to me not because the files were out-of-sync, but because one transitive dep failed to install. So this can happen sometimes not related to top-level deps. FYI

lib/parsers/package-lock-parser.ts
```diff
@@ -49,66 +64,240 @@ export class PackageLockParser implements LockfileParser {
...
+          name: depName,
+          requires: [],
+          version: dep.version,
+        };
+
+        if (dep.requires) {
+          depNode.requires = Object.keys(dep.requires);
+        }
+
+        const depKey: string[] = [...path, depName];
```
michael-go: would call this `depPath`/`depPathArray`, and `depKey` would be the `join`-ed string

lib/parsers/package-lock-parser.ts
```diff
@@ -261,11 +294,18 @@ export class PackageLockParser implements LockfileParser {
...
+        dependencies: dep.dependencies,
+        hasDevDependencies: dep.hasDevDependencies,
+        name: dep.name,
+        version: dep.version,
+      };
+      depTrees[depKey] = pkgTree;
       // Since this code doesn't handle any I/O or network, we need to force
       // event loop to tick while being used in server for request processing
-      await setImmediatePromise();
+      await this.setImmediatePromise();
```
michael-go: it's a minor thing, but would personally prefer the `setImmediatePromise` to be a static or even global function as it doesn't use any internal state of the the class instance.

lib/parsers/package-lock-parser.ts
```diff
@@ -128,63 +141,50 @@ export class PackageLockParser implements LockfileParser {
...
-      depGraph.setNode(newNode);
-      walkCycleRec(start, [], cycle, newNode);
+      const clonedNode = this.cloneNodeWithoutEdges(start, depMap, depGraph);
+      cycleStarts[start] = clonedNode;
+
+      /* FUNCTION DEFINITION
+        To keep an order of algorithm steps readable, function is defined on-the-fly
+        Arrow function is used for calling `this` without .bind(this) in the end
+      */
+      const acyclicDuplicationRec = (node, traversed, currentCycle, nodeCopy) => {
```
michael-go: this is great - and it will be even more readable if this definition moves 2 lines up, outside of the `for` loop. This way the algorithm can be read more outside-in.

lib/parsers/package-lock-parser.ts
```diff
@@ -261,11 +294,18 @@ export class PackageLockParser implements LockfileParser {
         const subDep = depTrees[subDepPath];
         dep.dependencies[subDep.name] = subDep;
       }
-      delete dep.requires;
-      depTrees[depKey] = {...dep as PkgTree};
+      const pkgTree: PkgTree = {
+        cyclic: dep.cyclic,
```
michael-go: FYI I see tests fail as `cyclic` & `hasDevDependencies` are always set although they are optional and sometimes not defined on `dep`.

miiila:
@michael-go I did my best to add a test with self-reference cycle. This was so far the best I was able to create, because I'm not aware about any existing package, which would transitively require itself (`trucolor` doesn't do it anymore, it transitively requires older version of itself 😂); since it has an impact on how packages are hoisted, it's not easy to manually adjuste `package-lock.json` to simulate this situation. 

So I created that kind of package by myself and algorithm doesn't break - it identifies cycle one level lower than it really is, but that's the limitation I'm happy to live with.

PR #22, opened 2018-10-02
chore: updated typescript template

Kirill89:
- [ ] Tests written and linted [ℹ︎]()

### What this does

updated typescript template to completely support node 4

### Notes for the reviewer

`String.prototype.endsWith()` not supported by es5.


miiila:
@Kirill89 Looks good, just a one question - `String.prototype.endsWith()` is not supported in `es5`, but is supported for Node 4 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith).

I wonder what right setup is here... even after reading lot of things, I'm still unsure :-D

PR #23, opened 2018-10-05
fix: Allow package.json to have missing dependency property

lili2311:

### What this does

Even though `dependencies` is a required field for npm, `npm init` doesn't seem to generate this property and users can easily not add the `dependencies` to the file. Return empty tree if this happens and do not error.
### Notes for the reviewer

### More information

- [SC-6322](https://snyksec.atlassian.net/browse/SC-6322)

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -47,9 +47,7 @@ async function buildDepTree(
   }
 
   const manifestFile: ManifestFile = parseManifestFile(manifestFileContents);
-  if (!manifestFile.dependencies && !includeDev) {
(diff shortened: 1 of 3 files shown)


adrukh:
Can you also add a test case where the `package.json` file simply doesn't have the `dependencies` key?

lili2311:
@adrukh That is exactly what this test is, it re-uses a fixture we already have which only has `devDependencies` https://github.com/snyk/nodejs-lockfile-parser/blob/master/test/lib/fixtures/dev-deps-only/package.json

PR #26, opened 2018-11-13
fix: skip file as a version key

miiila:

### What this does

Dependency can be specified as semver compatible string, git link oreven as a file path. In that case, dependency doesn't appear in package-lock / yarn.lock can't be looked up.

### More information
- [Internal customer messaging](https://app.hubspot.com/contacts/1699665/ticket/2127645)


lib/parsers/index.ts
```diff
@@ -68,6 +68,9 @@ export function getTopLevelDeps(targetFile: ManifestFile, includeDev: boolean):
   });
 
   for (const [name, version] of dependenciesIterator) {
+    if (/^file:/.test(version)) {
+      continue;
```
lili2311: Could we not keep it but not show the transitive `dependencies`? So we can say yep we saw this dep but we can't peek inside right now so it would be top level dep only with version exactly as it was given to us. Seems nicer than throwing it away

miiila: Very good point, I changed the approach, that makes more sense :-)

lili2311:
👏 nice!

PR #27, opened 2018-11-26
Feat/dont fail out of sync

lili2311:
- [ ] Tests written and linted [ℹ︎]()
- [ ] Commit history is tidy [ℹ︎]()

### What this does

_Explain why this PR exists_

### Notes for the reviewer

### More information

- [SC-XXXX]()
- [Link to documentation]()

### Screenshots

_Visuals that may help the reviewer_


lib/parsers/package-lock-parser.ts
```diff
@@ -119,7 +120,12 @@ export class PackageLockParser implements LockfileParser {
       } else if (/^file:/.test(dep.version)) {
         depTree.dependencies[dep.name] = createPkgTreeFromDep(dep);
       } else {
-        throw new OutOfSyncError(depName, 'npm');
+        if (!strictOutOfSync) {
```
miiila: I would sugget to flip the direction, i.e.:
```
if (strictOutOfSync) {
  throw new OutOfSyncError(depName, 'npm');
}
...
```

Saving one `else` statement :-)

test/lib/fixtures/out-of-sync-tree/expected-tree.json
```diff
@@ -0,0 +1,192 @@
...
+      },
+      "name": "debug",
+      "version": "4.1.0"
+    },
+    "body-parser": {
+      "depType": "prod",
+      "missingLockFileEntry": true,
+      "dependencies": {},
+      "name": "body-parser",
+      "version": "^1.18.2"
```
lili2311: note the original version is kept as that is all we know at this stage.

test/lib/fixtures/out-of-sync-tree/expected-tree.json
```diff
@@ -0,0 +1,192 @@
...
+          "name": "ms",
+          "version": "2.1.1"
+        }
+      },
+      "name": "debug",
+      "version": "4.1.0"
+    },
+    "body-parser": {
+      "depType": "prod",
+      "missingLockFileEntry": true,
```
lili2311: all dependencies will need to be filtered out on the Registry side, we can also there have the logic of using the filtered out deps as a user message to show the user

lib/parsers/yarn-lock-parse.ts
```diff
@@ -100,18 +102,25 @@ export class YarnLockParser implements LockfileParser {
     };
 
     const depKey = `${searchedDep.name}@${searchedDep.version}`;
-
+    // tslint:disable
```
Kirill89: Why are we need this?

lib/parsers/package-lock-parser.ts
```diff
@@ -119,7 +120,13 @@ export class PackageLockParser implements LockfileParser {
...
         depTree.dependencies[dep.name] = createPkgTreeFromDep(dep);
       } else {
-        throw new OutOfSyncError(depName, 'npm');
+        // TODO: also check the package version
+        // for a stricter check
+        if (strictOutOfSync) {
+          throw new OutOfSyncError(depName, 'npm');
+        }
+        depTree.dependencies[dep.name] = createPkgTreeFromDep(dep);
+        depTree.dependencies[dep.name].missingLockFileEntry = true;
```
Kirill89: Let's maybe move setting up `missingLockFileEntry` flag to `createPkgTreeFromDep` function?

`createPkgTreeFromDep(dep: Dep)` => `createPkgTreeFromDep(dep: Dep, missingLockFileEntry: boolean = false)`

lib/parsers/yarn-lock-parse.ts
```diff
@@ -100,18 +102,25 @@ export class YarnLockParser implements LockfileParser {
...
     const dep = _.get(lockFile.object, depKey);
 
-    if (!dep) {
+    if (!dep && strictOutOfSync) {
```
Kirill89: If you don't mind can we keep `if (!dep) {` condition and check `if (strictOutOfSync)` inside?

```
if (!dep) {
      if (strictOutOfSync) {
           throw new OutOfSyncError(searchedDep.name, 'yarn');
      }
      depSubTree.version = searchedDep.version;
      depSubTree.missingLockFileEntry = true;
      return depSubTree;
}
```

Kirill89: Same for package-lock

lib/parsers/package-lock-parser.ts
Kirill89: @miiila I just don't like this sequence: call some function (`createPkgTreeFromDep`) => patch something in the result of the function `missingLockFileEntry = true`. It is hard to read and possibly dangerous (`Cannot set property 'missingLockFileEntry' of undefined`).

Though, you maybe right.

PR #29, opened 2018-12-02
feat: lockfile parser must always return `depTree.name`

yuliabaron:

### What this does
When package name is missing use manifest filename instead

### More information
https://snyksec.atlassian.net/browse/SC-6758### Screenshots

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -23,7 +23,7 @@ export {
 async function buildDepTree(
   manifestFileContents: string, lockFileContents: string,
   includeDev = false, lockfileType?: LockfileType,
-  strict: boolean = true): Promise<PkgTree> {
+  strict: boolean = true, manifestFilePath: string = 'package.json'): Promise<PkgTree> {
 
   if (!lockfileType) {
     lockfileType = LockfileType.npm;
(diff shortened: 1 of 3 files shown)


lib/parsers/index.ts
```diff
@@ -52,9 +52,15 @@ export interface LockfileParser {
...
 export type Lockfile = PackageLock | YarnLock;
 
-export function parseManifestFile(manifestFileContents: string): ManifestFile {
+export function parseManifestFile(manifestFileContents: string, manifestFileName: string): ManifestFile {
   try {
-    return JSON.parse(manifestFileContents);
+    const manifestFile = JSON.parse(manifestFileContents);
+    // Fallback to the file name if name is not set
+    if (!manifestFile.name) {
+        manifestFile.name = manifestFileName;
```
Kirill89: Can I suggest to move all logic relevant to `manifestFileName` to `/lib/index.ts`? Looks like it is not related to parsers because they only need to parse existing file. And also it is less changes (do not need to pass `manifestFileName` everywhere).

lib/index.ts
```diff
@@ -21,7 +21,7 @@ export {
 };
 
 async function buildDepTree(
-  manifestFileContents: string, lockFileContents: string,
+  manifestFileContents: string, manifestFileName: string, lockFileContents: string,
```
Kirill89: Let's maybe make it last argument and make it optional? This way you'll keep API consistent.

PR #33, opened 2018-12-18
feat: add tree size in the output

Kirill89:

### What this does

We want to know how many dependencies totally in the tree. It's useful for debug purposes when, for example, we want to measure why so much time we spend on parsing.

### DIFF
--- a/lib/parsers/package-lock-parser.ts
+++ b/lib/parsers/package-lock-parser.ts
@@ -75,6 +75,7 @@ export class PackageLockParser implements LockfileParser {
       dependencies: {},
       hasDevDependencies: !_.isEmpty(manifestFile.devDependencies),
       name: manifestFile.name,
+      size: 1,
       version: manifestFile.version || '',
     };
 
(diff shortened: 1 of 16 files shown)


lib/parsers/package-lock-parser.ts
```diff
@@ -75,6 +75,7 @@ export class PackageLockParser implements LockfileParser {
       dependencies: {},
       hasDevDependencies: !_.isEmpty(manifestFile.devDependencies),
       name: manifestFile.name,
+      size: 1,
```
miiila: This doesn't have to be true because of https://github.com/snyk/nodejs-lockfile-parser/pull/33/files#diff-e2f9058bd2d55f7bf1abe78471eec15aR83

lib/parsers/package-lock-parser.ts
```diff
@@ -296,12 +308,14 @@ export class PackageLockParser implements LockfileParser {
...
       const depKey = depOrder.shift() as string;
       const dep = depMap[depKey];
+      let treeSize = 1;
 
       // direction is from the dependency to the package requiring it, so we are
       // looking for predecessors
       for (const subDepPath of depGraph.predecessors(depKey)) {
         const subDep = depTrees[subDepPath];
         dep.dependencies[subDep.name] = subDep;
+        treeSize += depTreesSizes[subDepPath];
```
miiila: I have no problem with this approach, but you can do it completely without `depTreesSizes`. I believe `dep.size += subDep.size + 1;` in every iteration is all you need, since you are builduing it gradually bottom-up. And every `subDep` is an instance of `PkgTree`, so it has `size available`.

lib/parsers/package-lock-parser.ts
Kirill89: It's exactly why it's here. We always have root dependency. So, minimum value is 1.

lib/parsers/package-lock-parser.ts
Kirill89: In this case each node will have size. But I want it to be only in the root node.

PR #38, opened 2019-02-04
feat: Allow us to identify yarn workspaces in package.json

lili2311:

### What this does
Identify if a `package.json` is a yarn workspace, we can start using this to log/treat/process these projects differenly

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -2,8 +2,16 @@ import 'source-map-support/register';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as _ from 'lodash';
-import {LockfileParser, Lockfile, ManifestFile, PkgTree,
-  DepType, parseManifestFile, LockfileType} from './parsers';
+import {
+  LockfileParser,
+  Lockfile,
+  ManifestFile,
+  PkgTree,
+  DepType,
+  parseManifestFile,
+  LockfileType,
+  getYarnWorkspaces,
+} from './parsers';
 import {PackageLockParser} from './parsers/package-lock-parser';
 import {YarnLockParser} from './parsers/yarn-lock-parse';
 import getRuntimeVersion from './get-node-runtime-version';
@@ -12,6 +20,8 @@ import {UnsupportedRuntimeError, InvalidUserInputError, OutOfSyncError} from './
 export {
   buildDepTree,
   buildDepTreeFromFiles,
+  getYarnWorkspacesFromFiles,
+  getYarnWorkspaces,
   PkgTree,
   DepType,
   LockfileType,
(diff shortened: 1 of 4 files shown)


lib/index.ts
```diff
@@ -2,16 +2,28 @@ import 'source-map-support/register';
...
-  DepType, parseManifestFile, LockfileType} from './parsers';
+import {
+  LockfileParser,
+  Lockfile,
+  ManifestFile,
+  PkgTree,
+  DepType,
+  parseManifestFile,
+  LockfileType,
+  getYarnWorkSpaces,
```
miiila: super NIT: `WorkSpace` -> `Workspace`

lib/parsers/index.ts
```diff
@@ -92,3 +99,22 @@ export function createPkgTreeFromDep(dep: Dep): PkgTree {
 
   return pkgTree;
 }
+
+export function getYarnWorkSpaces(targetFile: string): YarnWorkSpaceResponse {
```
miiila: I'm not sure we need `isYarnWorkSpace`, can't we just check `workspaces.length`?
```suggestion
export function getYarnWorkSpaces(targetFile: string): YarnWorkSpaceResponse {
  try {
    const packageJson: ManifestFile = parseManifestFile(targetFile);
    const workspaces: string[] = !!packageJson.workspaces && !!packageJson.private ? ...packageJson.workspaces : [];
    return workspaces;
  } catch (e) {
    throw new InvalidUserInputError('package.json parsing failed with ' +
      `error ${e.message}`);
  }
}
```

lili2311:
> Do I understand correctly this is just a preparation and we will use this in a future? Otherwise, looks 🆗 , just consider my suggestion about dropping `isYarnWorkspace`

Indeed it is just prep, we will have to do this check on every yarn project to identify if it is a workspace. It felt wrong just returning empty workspaces because it then did not give me a clue if this project is actually a workspace project but with empty list or if it is not a workspace project.

A workspace project will always have lockfile in the root, others will have it same dir as manifest.

What would you suggest? I can do something that returns different type like False or [] to differentiate?

lib/parsers/index.ts
```diff
@@ -92,3 +94,16 @@ export function createPkgTreeFromDep(dep: Dep): PkgTree {
 
   return pkgTree;
 }
+
+export function getYarnWorkspaces(targetFile: string): string [] | false {
+  try {
+    const packageJson: ManifestFile = parseManifestFile(targetFile);
+    if (!!packageJson.workspaces && !!packageJson.private) {
+      return packageJson.workspaces;
```
miiila: Would you mind to `return ...packageJson.workspaces;`? By that, you will copy the values from the original object rather than keeping a reference to it. So we will prevent unlikely, but possible issues with weird missing data.

PR #45, opened 2019-07-15
Feat: move dep type to label

lwywoo:
- [ ] Documentation written / README.md updated [https://snyk.io/docs/snyk-for-dotnet/](i)
- [ ] Reviewed by Snyk team

### What this does
Move property `depType` to be nested in `label` property to be used in depGraph conversion.

### More information
- https://snyksec.atlassian.net/browse/BST-753

### DIFF
--- a/lib/parsers/index.ts
+++ b/lib/parsers/index.ts
@@ -34,7 +34,9 @@ export interface PkgTree {
   meta?: {
     nodeVersion: string;
   };
-  depType?: DepType;
+  labels?: {
+    depType: DepType;
+  };
   hasDevDependencies?: boolean;
   cyclic?: boolean;
   missingLockFileEntry?: boolean;
@@ -92,8 +94,10 @@ export function getTopLevelDeps(targetFile: ManifestFile, includeDev: boolean): 
 
 export function createPkgTreeFromDep(dep: Dep): PkgTree {
   const pkgTree: PkgTree = {
-    depType: dep.dev ? DepType.dev : DepType.prod,
     dependencies: {},
+    labels: {
+      depType: dep.dev ? DepType.dev : DepType.prod,
+    },
     name: dep.name,
     version: dep.version,
   };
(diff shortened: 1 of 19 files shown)


test/lib/package-lock.test.ts
```diff
@@ -116,6 +116,18 @@ test('Parse npm package-lock.json with dev deps only', async (t) => {
...
+test('Parse npm simple package-lock.json dev and prod deps', async (t) => {
+  const expectedDepTree = load('simple-test/expected-tree.json');
+  const depTree = await buildDepTreeFromFiles(
+    `${__dirname}/fixtures/simple-test/`,
+    'package.json',
+    'package-lock.json',
+    true,
+  );
+
+  t.deepEqual(depTree, expectedDepTree, 'Tree is created with dev deps only');
```
lili2311: `dev and prod deps` so should the message here also reflect that it is expected to have both?

lib/parsers/index.ts
```diff
@@ -34,7 +34,9 @@ export interface PkgTree {
   meta?: {
     nodeVersion: string;
   };
-  depType?: DepType;
+  labels?: {
+    depType: DepType;
+  };
   hasDevDependencies?: boolean;
   cyclic?: boolean;
   missingLockFileEntry?: boolean;
```
lili2311: are these already preserved btw? 
- `cyclic`
- `missingLockFileEntry`
- `cyclic`

test/lib/package-lock.test.ts
lwywoo: Updated

lib/parsers/index.ts
lwywoo: - Cycles are not yet supported in depGraph: https://github.com/snyk/dep-graph/blob/master/src/legacy/index.ts#L221-L224
- `missingLockFileEntry` - don't think this is preserved either

PR #51, opened 2019-10-18
Fix/align deptree interface

miiila:
- [ ] Documentation written / README.md updated [https://snyk.io/docs/snyk-for-dotnet/](i)
- [x] Reviewed by Snyk team

### What this does

DepTree intreface has diverged from the one used in e.g. DepGraph.
The biggest difference is the dependencies keys, which is not mandatory anymore.
This change is a prereq for using depgraphs.

### Notes for the reviewer

Just carefully review tests and think of potential impact and omitted parts.

### More information

- [BST-631](https://snyksec.atlassian.net/browse/BST-631)

### DIFF
--- a/lib/parsers/index.ts
+++ b/lib/parsers/index.ts
@@ -28,7 +28,7 @@ export interface ManifestFile {
 export interface PkgTree {
   name: string;
   version: string;
-  dependencies: {
+  dependencies?: {
     [dep: string]: PkgTree;
   };
   meta?: {
@@ -37,10 +37,10 @@ export interface PkgTree {
   labels?: {
      scope?: Scope;
      pruned?: 'cyclic';
+     missingLockFileEntry?: boolean;
   };
   hasDevDependencies?: boolean;
   cyclic?: boolean;
-  missingLockFileEntry?: boolean;
   size?: number;
 }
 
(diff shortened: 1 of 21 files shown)


lili2311:
👋 Please could you give a little more context:
- Why does removed empty dependencies object helping towards moving to a graph?
- How does Registry and other upstreams handle this change? Do they care?

miiila:
@lili2311 

> Why does removed empty dependencies object helping towards moving to a graph?

As I say in description, `DepTree intreface has diverged from the one used in e.g. DepGraph.` The main difference is in mandatory vs optional `dependencies` property. Once moved to `graph function`, tree created out of the graph should be the same as the one created by `tree function`. Therefore the interface has to be same and that's why we need to align those.

> How does Registry and other upstreams handle this change? Do they care?

That's what we spoke about with @darscan. What needs to be put back is top level `dependencies` key which might be somewhere used to checks. Apart from that, everything should be ok.

darscan:
> What needs to be put back is top level dependencies key which might be somewhere used to checks

Can we do that then please :D

darscan:
I still think you need to split the type in two, like we do in the dep-graph lib: top-level tree type (with mandatory `dependencies` property) VS recursive dep type (with optional `dependencies` property)

Or, just use the dep-graph type in

PR #59, opened 2020-04-19
feat: apply limit on tree size.

anthogez:
- [x] Reviewed by Snyk team

Applies limit on tree size

### DIFF
--- a/lib/parsers/package-lock-parser.ts
+++ b/lib/parsers/package-lock-parser.ts
@@ -7,7 +7,7 @@ import {
   LockfileParser, PkgTree, DepTreeDep, Dep, Scope, ManifestFile,
   getTopLevelDeps, Lockfile, LockfileType, createDepTreeDepFromDep,
 } from './';
-import {InvalidUserInputError, OutOfSyncError} from '../errors';
+import {InvalidUserInputError, OutOfSyncError, TreeSizeLimitError} from '../errors';
 
 export interface PackageLock {
   name: string;
@@ -120,8 +120,12 @@ export class PackageLockParser implements LockfileParser {
 
     // number of dependencies including root one
     let treeSize = 1;
-
+    const treeSizeLimit = 6.0e6;
     for (const dep of topLevelDeps) {
+      // tree size limit should be 6 millions.
+      if (treeSize > treeSizeLimit) {
+        throw new TreeSizeLimitError();
+      }
       // if any of top level dependencies is a part of cycle
       // it now has a different item in the map
       const depName = cycleStarts[dep.name] || dep.name;
--- a/lib/errors/index.ts
+++ b/lib/errors/index.ts
@@ -1,3 +1,4 @@
 export {InvalidUserInputError} from './invalid-user-input-error';
 export {UnsupportedRuntimeError} from './unsupported-runtime-error';
 export {OutOfSyncError} from './out-of-sync-error';
+export {TreeSizeLimitError} from './tree-size-limit-error';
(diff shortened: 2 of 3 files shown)


lib/parsers/package-lock-parser.ts
```diff
@@ -120,8 +120,13 @@ export class PackageLockParser implements LockfileParser {
 
     // number of dependencies including root one
     let treeSize = 1;
-
+    const treeSizeLimit = 6.0e6;
+  
     for (const dep of topLevelDeps) {
+      // On Call we decided that the tree size limit will be of 6 millions.
```
dkontorovskyy, line 126: Probably this shouldn't be there. Just give brief explanation, please

orsagie, line 126: This might be better next to the definition of the const
```suggestion
      // oncall decided tree size limit should be 6 million
```

dkontorovskyy: Shall we have a test for this?

lib/parsers/package-lock-parser.ts
```diff
@@ -120,8 +120,13 @@ export class PackageLockParser implements LockfileParser {
 
     // number of dependencies including root one
     let treeSize = 1;
-
+    const treeSizeLimit = 6.0e6;
+  
     for (const dep of topLevelDeps) {
+      // Tree size limit will be of 6 millions.
```
dkontorovskyy, line 126: Shall we have a test for this?

lib/parsers/package-lock-parser.ts
```diff
@@ -120,8 +120,13 @@ export class PackageLockParser implements LockfileParser {
 
     // number of dependencies including root one
     let treeSize = 1;
-
+    const treeSizeLimit = 6.0e6;
+  
     for (const dep of topLevelDeps) {
+      // tree size limit should be 6 millions.
+      if (treeSize > treeSizeLimit) {
+        throw new Error('Tree size has surpassed current tree size limit.');
```
lili2311, line 127: Can we please make this a named error like `OutOfSyncError` so we can catch it upstream and present it nicely?

lib/parsers/package-lock-parser.ts
anthogez: We can have a side pr with fixtures where we can simulate the 6m+ test

dkontorovskyy: Why can't it be here?

lib/parsers/package-lock-parser.ts
```diff
@@ -120,8 +120,12 @@ export class PackageLockParser implements LockfileParser {
 
     // number of dependencies including root one
     let treeSize = 1;
-
+    const treeSizeLimit = 6.0e6;
     for (const dep of topLevelDeps) {
+      // tree size limit should be 6 millions.
+      if (treeSize > treeSizeLimit) {
+        throw new OutOfSyncError('Tree size has surpassed current tree size limit.', 'npm');
```
dkontorovskyy, line 127: This is not OutOfSyncError. Please, create new `named` error, like the OutOfSyncError and apply here, please

lib/parsers/package-lock-parser.ts
anthogez: I wanted to apply that in side pr, due of the current needs. But no worries I will update with test.

anthogez: Due of current priority, we do not have time to create a fixture for that now.

PR #63, opened 2020-04-22
feat: config tree size limit for yarn and npm.

anthogez:
- [x] Reviewed by Snyk team

### What this does
config treeSizeLimit for yarn & npm

This will be improved in another pr with `configurable treeSizeLimit` tests for yarn & npm

Edit --
Side pr no longer necessary because the side pr upgrade are already applied on master so we could proceed with this

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,0 +1,2 @@
+import { loadConfig } from 'snyk-config';
+export const config: any = loadConfig(__dirname + '../..');
--- a/config.default.json
+++ b/config.default.json
@@ -1,0 +1,4 @@
+{
+  "NPM_TREE_SIZE_LIMIT": 6.0e6,
+  "YARN_TREE_SIZE_LIMIT": 6.0e6
+}
--- a/lib/parsers/package-lock-parser.ts
+++ b/lib/parsers/package-lock-parser.ts
@@ -1,6 +1,7 @@
 import * as _ from 'lodash';
 import * as graphlib from 'graphlib';
 import * as uuid from 'uuid/v4';
+import { config } from '../config';
 import { setImmediatePromise } from '../set-immediate-promise';
 
 import {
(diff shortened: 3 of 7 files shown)


lib/config.ts
```diff
@@ -0,0 +1,5 @@
+const config = {
+    NPM_TREE_SIZE_LIMIT: 6.0e6,
+    YARN_TREE_SIZE_LIMIT: 6.0e6,
+};
+export const {NPM_TREE_SIZE_LIMIT, YARN_TREE_SIZE_LIMIT} = config;
```
gitphill, line 5: we should use our config library for this https://github.com/snyk/config

anthogez: We cannot do that yet, requires further work in terms of `ts versioning` and I prefer create a separated pr for that

anthogez:
> can we add a test to assert a TreeSizeImitError?

Will be added with the new config support

lib/config.ts
dkontorovskyy: We should use our config library in here, rater that adding more configs. What's the issue with `ts versioning` you mentioned above?

anthogez: This project in terms of config is still behind others and need certains upgrade that I prefer avoid on this pr since that they are not related.

anthogez: As I promised I created a side pr that solve the concerns that I had before of implement our config lib, now I have update this pr ⚡

anthogez: As I promised I created a side pr that solve the concerns that I had before of implement our config lib, now I have update this pr ⚡freely :)

anthogez:
> can we add a test to assert a TreeSizeImitError?

Now that snyk-config was implemented and we will no longer need a side pr, now we have as well the tests.

anthogez:
> Looks good to me, can there be any tests in place for that?

Were already added :)

PR #77, opened 2020-07-02
fix: use filename as default name istead of full path

dkontorovskyy:
### What this does
Use filename as default name instead of the full path

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -76,7 +76,9 @@ async function buildDepTree(
 
   const manifestFile: ManifestFile = parseManifestFile(manifestFileContents);
   if (!manifestFile.name) {
-    manifestFile.name = defaultManifestFileName;
+    manifestFile.name = path.isAbsolute(defaultManifestFileName)
+      ? path.basename(defaultManifestFileName)
+      : defaultManifestFileName;
   }
 
   const lockFile: Lockfile = lockfileParser.parseLockFile(lockFileContents);
--- a/test/lib/index.test.ts
+++ b/test/lib/index.test.ts
@@ -1,0 +1,83 @@
+import { test } from 'tap';
+import * as path from 'path';
+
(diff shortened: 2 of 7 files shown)


test/lib/index.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    'package-json-without-name/package.json',
+  );
+  const lockFileContent: string = await readFixture(
+    'package-json-without-name/package-lock.json',
+  );
+  const absolutePath = path.resolve(
+    __dirname,
+    'lib',
+    'fixtures',
+    'package.json',
```
orsagie, line 41: can you change this to something else to show that this absolutePath is in fact being used properly. (currently both tests have the exact same assert, which makes it hard to see what the difference between both calls is)

lib/index.ts
```diff
@@ -76,7 +76,7 @@ async function buildDepTree(
 
   const manifestFile: ManifestFile = parseManifestFile(manifestFileContents);
   if (!manifestFile.name) {
-    manifestFile.name = defaultManifestFileName;
+    manifestFile.name = path.basename(defaultManifestFileName);
```
orsagie, line 79: how is changing the name going to affect existing projects?

test/lib/index.test.ts
dkontorovskyy: The difference is in incoming parameters. That output should be the same. That is why I have same asserts :)

lib/index.ts
dkontorovskyy: Good catch! Fixed

dkontorovskyy:
@orsagie, in yarnv2 for CLI we will be calling `buildDepTree` instead of `buildDepTreeForFiles`, because of that we are gonna be passing default name from CLI. This gonna brake yarnWorkspaces functionality. With this small fix, it will act as it used to be.

PR #86, opened 2020-09-03
fix: update dep-graph to use upstream lodash

JackuB:
Updating dep-graph to prevent including our lodash fork.

### DIFF
--- a/lib/parsers/package-lock-parser.ts
+++ b/lib/parsers/package-lock-parser.ts
@@ -2,7 +2,7 @@ import * as _cloneDeep from 'lodash.clonedeep';
 import * as _isEmpty from 'lodash.isempty';
 import * as _set from 'lodash.set';
 import * as _toPairs from 'lodash.topairs';
-import * as graphlib from '@snyk/graphlib';
+import * as graphlib from 'graphlib';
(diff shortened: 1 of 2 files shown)


lib/parsers/package-lock-parser.ts
```diff
@@ -2,7 +2,7 @@ import * as _cloneDeep from 'lodash.clonedeep';
 import * as _isEmpty from 'lodash.isempty';
 import * as _set from 'lodash.set';
 import * as _toPairs from 'lodash.topairs';
-import * as graphlib from '@snyk/graphlib';
+import * as graphlib from 'graphlib';
```
aviadhahami, line 5: did we omit the snyk prefix from the graphlib?

aviadhahami: https://github.com/snyk/graphlib/blob/master/package.json#L2 :thinking:

JackuB: We don't want to use our fork anymore - it was only swapping a dependency for our lodash fork https://github.com/dagrejs/graphlib/compare/master...snyk:master

aviadhahami (approved):

PR #92, opened 2021-02-22
feat: support dep graph

admons:
- [ ] Reviewed by Snyk team

### What this does

Adds support for parsing `package-lock` and `yarn.lock` into [DepGraph](https://github.com/snyk/dep-graph)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 .eslintcache
 *.log
 /package-lock.json
+.idea
--- a/lib/common.ts
+++ b/lib/common.ts
@@ -1,0 +1,137 @@
+import * as path from 'path';
+import * as fs from 'fs';
+import { InvalidUserInputError, UnsupportedError } from './errors';
+import {
+  Lockfile,
+  LockfileParser,
+  LockfileType,
+  ManifestFile,
+  parseManifestFile,
+} from './parsers';
+import { PackageLockParser } from './parsers/package-lock-parser';
+import { YarnLockParser } from './parsers/yarn-lock-parse';
+
+export function parseManifestAndLock(
+  manifestFileContents: string,
+  lockFileContents: string,
+  lockfileParser: LockfileParser,
+  defaultManifestFileName: string = 'package.json',
+): {
+  manifestFile: ManifestFile;
+  lockFile: Lockfile;
+} {
+  const manifestFile: ManifestFile = parseManifestFile(manifestFileContents);
+  if (!manifestFile.name) {
+    manifestFile.name = path.isAbsolute(defaultManifestFileName)
+      ? path.basename(defaultManifestFileName)
+      : defaultManifestFileName;
+  }
+
+  const lockFile: Lockfile = lockfileParser.parseLockFile(lockFileContents);
+  return { manifestFile, lockFile };
+}
+
+export function getLockFileParser(
+  lockFileContents: string,
+  lockfileType?: LockfileType,
+): LockfileParser {
+  if (!lockfileType) {
+    lockfileType = LockfileType.npm;
+  } else if (lockfileType === LockfileType.yarn) {
+    lockfileType = getYarnLockfileType(lockFileContents);
+  }
+
+  switch (lockfileType) {
+    case LockfileType.npm:
+      return new PackageLockParser();
+    case LockfileType.yarn:
+      return new YarnLockParser();
+    case LockfileType.yarn2:
+      throw new UnsupportedError(
+        'Yarn2 support has been temporarily removed to support Node.js versions 8.x.x',
+      );
+    /**
+     * Removing yarn 2 support as this breaks support for yarn with Node.js 8
+     * See: https://github.com/snyk/snyk/issues/1270
+     *
+     * Uncomment following code once Snyk stops Node.js 8 support
+     * // parsing yarn.lock is supported for Node.js v10 and higher
+     * if (getRuntimeVersion() >= 10) {
+     *  lockfileParser = new Yarn2LockParser();
+     *  } else {
+     *   throw new UnsupportedRuntimeError(
+     *     'Parsing `yarn.lock` is not ' +
+     *       'supported on Node.js version less than 10. Please upgrade your ' +
+     *       'Node.js environment or use `package-lock.json`',
+     *   );
+     * }
+     * break;
+     */
+    default:
+      throw new InvalidUserInputError(
+        'Unsupported lockfile type ' +
+          `${lockfileType} provided. Only 'npm' or 'yarn' is currently ` +
+          'supported.',
+      );
+  }
+}
+
+export function getFilesContent(
+  root: string,
+  manifestFilePath: string,
+  lockFilePath: string,
+) {
+  if (!root || !manifestFilePath || !lockFilePath) {
+    throw new Error('Missing required parameters for buildDepTreeFromFiles()');
+  }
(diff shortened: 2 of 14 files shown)


lib/common.ts
```diff
@@ -0,0 +1,146 @@
...
+  Lockfile,
+  LockfileParser,
+  LockfileType,
+  ManifestFile,
+  parseManifestFile,
+} from './parsers';
+import { PackageLockParser } from './parsers/package-lock-parser';
+import { YarnLockParser } from './parsers/yarn-lock-parse';
+
+export type BuildFunc = (
```
admons, line 14: remove

lib/parsers/get-package-lock-dep-graph.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  const topLevelDeps: Dep[] = getTopLevelDeps(manifestFile, includeDev);
+
+  const devMemoizationMap: Map<PackageLockDep, NodeId> = new Map();
+  const prodMemoizationMap: Map<PackageLockDep, NodeId> = new Map();
+
+  for (const dep of topLevelDeps) {
+    const dependency = packageLock.dependencies?.[dep.name];
+    if (/^file:/.test(dep.version)) {
```
aviadhahami, line 62: This is a bit confusing and can make one thing (like myself) that we may see an infinite loop here.

Since we're already iterating top-level-deps, what is the meaning of adding another top-level dep?

lib/parsers/get-package-lock-dep-graph.ts
```diff
@@ -0,0 +1,167 @@
...
+function find(key: string, ancestors: PackageLockDep[] = []): PackageLockDep {
+  for (const dep of ancestors) {
+    if (dep.dependencies?.[key]) {
+      return dep.dependencies[key];
+    }
+  }
+  throw new Error('');
+}
+
+type NodeId = string;
```
aviadhahami, line 128: Should this be exported or moved to `types`?

lib/parsers/get-package-lock-dep-graph.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+function find(key: string, ancestors: PackageLockDep[] = []): PackageLockDep {
+  for (const dep of ancestors) {
+    if (dep.dependencies?.[key]) {
+      return dep.dependencies[key];
+    }
+  }
+  throw new Error('');
```
aviadhahami, line 125: Which error should we throw in this case?

lib/parsers/get-package-lock-dep-graph.ts
```diff
@@ -0,0 +1,167 @@
...
+      dependency?.linkedRequires,
+      isDev,
+      isDev ? devMemoizationMap : prodMemoizationMap,
+      nodeId,
+    );
+  }
+
+  return depGraphBuilder.build();
+
+  function addTopLevelDep(dep: Dep, version: string = dep.version) {
```
aviadhahami, line 85: Can we extract this function from `getPackageLockDepGraph`?

lib/parsers/get-package-lock-dep-graph.ts
```diff
@@ -0,0 +1,167 @@
...
+    depGraphBuilder.addPkgNode({ name: dep.name, version }, nodeId, {
+      labels: { scope: dep.dev ? Scope.dev : Scope.prod },
+    });
+    depGraphBuilder.connectDep(depGraphBuilder.rootNodeId, nodeId);
+    return nodeId;
+  }
+}
+
+function getNodeId(depName: string, version: string = '') {
+  return `${depName}@${version}|${Math.random()}`;
```
aviadhahami, line 96: Are we okay w/ having the leading `0.` from the random call?

lib/parsers/get-yarn-lock-dep-graph.ts
```diff
@@ -0,0 +1,150 @@
...
+import * as pMap from 'p-map';
+import { YarnLock } from './yarn-lock-parse';
+import { eventLoopSpinner } from 'event-loop-spinner';
+import { OutOfSyncError } from '../errors';
+import {
+  EVENT_PROCESSING_CONCURRENCY,
+  YarnLockDep,
+} from './yarn-lock-parse-base';
+
+type VersionHolder = YarnLockDep | { version: string };
```
aviadhahami, line 12: let's move types to a unified place

lib/parsers/get-yarn-lock-dep-graph.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const pkgManager = {
+    name: 'yarn',
+    version: yarnLock.type === LockfileType.yarn ? '1' : '2',
+  };
+  const depGraphBuilder = new DepGraphBuilder(pkgManager, rootPkgInfo);
+  const prodMemoizationMap = new Map<VersionHolder, string>();
+  const devMemoizationMap = new Map<VersionHolder, string>();
+
+  // const nodeVersion = manifestFile?.engines?.node;
```
aviadhahami, line 39: #cleanme

aviadhahami (commented):
Please add JSdoc for methods as some of the logics here are not easy to understand at a glance.

Also - can we please have unit-tests for the functions? so we can `f(x)=y` for future debugging? 

Overall looking good :) [and not simple]

.gitignore
```diff
@@ -6,3 +6,4 @@ dist
 .eslintcache
 *.log
 /package-lock.json
+.idea
```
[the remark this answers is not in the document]
admons: I tried it couple of times and it doesn't work.
I think it's ok that our `.gitignore` files will have this line
Why do you even care?????????????????????????

lib/dep-graph.ts
```diff
@@ -0,0 +1,49 @@
...
+  includeDev = false,
+  strict = true,
+): Promise<DepGraph> {
+  const {
+    manifestFileContents,
+    lockFileContents,
+    lockFileType,
+  } = getFilesContent(root, manifestFilePath, lockFilePath);
+
+  return await buildDepGraph(
```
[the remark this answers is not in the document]
admons: Having `await` before returning helps with the stack trace on errors

lib/parsers/get-package-lock-dep-graph.ts
admons: `addTopLevelDep` name is responsible for the confusion, so I renamed it to `addTopLevelDepNodeToGraph`

lib/parsers/get-package-lock-dep-graph.ts
admons: No, it's just an alias for internal types to use

lib/parsers/get-package-lock-dep-graph.ts
admons: Sorry, this was a WIP
I changed it to only add `|number` when there's more than node with the same node id

lib/parsers/get-yarn-lock-dep-graph.ts
admons: It's only for internal use in this file
There's no reason to create a centric types file for types that are not being used outside of a specific file

lib/parsers/get-yarn-lock-dep-graph.ts
admons: #ok

PR #93, opened 2021-02-25
feat: optimised yarn depTree

admons:
- [x] Tests written and linted
- [x] Documentation written / README.md updated [https://snyk.io/docs/snyk-for-node/](i)
- [x] Follows [CONTRIBUTING agreement](CONTRIBUTING.md)
- [x] Commit history is tidy [https://git-scm.com/book/en/v2/Git-Branching-Rebasing](i)
- [ ] Reviewed by Snyk team

### What this does

Creates an optimised dep-tree version for yarn by reusing package-lock logic
I extracted everything that is shared from `package-lock-parser` into `lock-parser-base`

The implications for [that ticket](https://snyk.zendesk.com/agent/tickets/3396) is ~220 million nodes for around 900 dependencies

Before this change, yarn projects looks like the right tree while npm projects looks like the left one
image

PR #94, opened 2021-03-01
BREAKING CHANGE: remove tree size limit

admons:
- [ ] Reviewed by Snyk team

### What this does

As we are creating optimised trees (see https://github.com/snyk/nodejs-lockfile-parser/pull/93), and that in most places we don't serialise the DepTree object, we don't need to really care about the size of the tree, but on the number of dependencies.
After reviewing most of the "big trees" tickets in Zendesk, the number of dependencies are limited by thousands, which shouldn't be a problem any more. 

This pr removes the limit of the tree size.

This is a breaking change, as we must verify that this plugin consumers doesn't serialise this object.

### DIFF
--- a/config.default.json
+++ b/config.default.json
@@ -1,4 +1,1 @@
-{
-  "NPM_TREE_SIZE_LIMIT": 6.0e6,
-  "YARN_TREE_SIZE_LIMIT": 6.0e6
-}
+{}
--- a/lib/parsers/lock-parser-base.ts
+++ b/lib/parsers/lock-parser-base.ts
@@ -18,11 +18,7 @@ import {
   PkgTree,
   Scope,
 } from './';
-import {
-  InvalidUserInputError,
-  OutOfSyncError,
-  TreeSizeLimitError,
-} from '../errors';
+import { InvalidUserInputError, OutOfSyncError } from '../errors';
 
 export interface PackageLockDeps {
   [depName: string]: PackageLockDep;
@@ -57,7 +53,7 @@ interface EdgeDirection {
 export abstract class LockParserBase implements LockfileParser {
   protected pathDelimiter = '|';
 
-  constructor(protected type: LockfileType, protected treeSizeLimit: number) {}
+  constructor(protected type: LockfileType) {}
 
   public abstract parseLockFile(lockFileContents: string): Lockfile;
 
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -139,16 +139,6 @@ workflows:
             branches:
               ignore:
                 - master
-      - test-windows:
```
admons, line 142: In addition, I also removed the node v8 tests as we don't support them anymore

anthogez:
Could you update also to a higher version? https://github.com/snyk/nodejs-lockfile-parser/blob/master/package.json#L25

.circleci/config.yml
```diff
@@ -139,16 +139,6 @@ workflows:
...
-          name: Windows Tests for Node v8 support
-          context: nodejs-install
-          node_version: "8.17.0"
```
anthogez, line 145: Regarding this, could you update also to a higher version? https://github.com/snyk/nodejs-lockfile-parser/blob/master/package.json#L25

robcresswell:
Can we break up PRs with separate concerns please? We shouldn't be changing CI config at the same time of config stuff. Makes rollbacks harder.

anthogez:
In addition to what was already mentioned about `separate prs`.
Would be nice to have the breaking change pr following this pattern

commit label:
refactor:  remove tree size limit

commit message body: (same content of your GH notes)
BREAKING CHANGE: As we are creating optimised trees (see #93), and that in most places we don't serialise the DepTree object, we don't need to really care about the size of the tree, but on the number of dependencies.
After reviewing most of the "big trees" tickets in Zendesk, the number of dependencies are limited by thousands, which shouldn't be a problem any more.

This pr removes the limit of the tree size.
This is a breaking change, as we must verify that this plugin consumers doesn't serialise this object.

.circleci/config.yml
admons: https://github.com/snyk/nodejs-lockfile-parser/pull/95

anthogez (approved):

mhassan1:
This was a blocker for me. As a workaround, I was able to execute `SNYK_YARN_TREE_SIZE_LIMIT=999999999999 snyk test` to ignore the limit.

Issue #99, opened 2021-04-06
yarn 2 `resolutions` causes `OutOfSyncError`


mhassan1:
As part of the optimization of the `depTree` in https://github.com/snyk/nodejs-lockfile-parser/pull/93, we added the following check: https://github.com/snyk/nodejs-lockfile-parser/blob/0bd6a0a2a8a629d9929d089d13cac7e4332e5970/lib/parsers/lock-parser-base.ts#L346-L348

In the case of `resolutions` in the `package.json` in Yarn 2, this error will get thrown because of the way Yarn 2 puts custom resolutions in `yarn.lock`.

**Reproduction Steps**

1. Create a new Yarn 2 project with `package.json`:
```json
{
  "dependencies": {
    "lodash": "4.17.19"
  },
  "resolutions": {
    "lodash": "4.17.20"
  }
}
```
2. Run `yarn`, which will create a `yarn.lock`:
```yaml
"lodash@npm:4.17.20":
  version: 4.17.20
  resolution: "lodash@npm:4.17.20"
  checksum: b31afa09739b7292a88ec49ffdb2fcaeb41f690def010f7a067eeedffece32da6b6847bfe4d38a77e6f41778b9b2bca75eeab91209936518173271f0b69376ea
  languageName: node
  linkType: hard

"root-workspace-0b6124@workspace:.":
  version: 0.0.0-use.local
  resolution: "root-workspace-0b6124@workspace:."
  dependencies:
    lodash: 4.17.19
  languageName: unknown
  linkType: soft
```
3. Run `snyk test`
4. See error `Dependency lodash@4.17.19 was not found in undefined. Your package.json and undefined are probably out of sync. Please run "undefined" and try again.` (the `undefined` is just because `yarn2` is missing from `lib/errors/out-of-sync-error.ts`.

**Explanation**
In the above `yarn.lock`, the root workspace has a dependency on `lodash@4.17.19`, which does not appear elsewhere in the `yarn.lock` because the only actual `lodash` is `4.17.20`, caused by the resolution. This breaks the `depMap` check, above, even when `--strict-out-of-sync=false` is passed.

**Potential Solutions**
1. When `--strict-out-of-sync=false` is passed, ignore dependencies that are not in the `depMap`; this will unblock consumers who use `resolutions`, but it doesn't add full support for `resolutions`.
2. Figure out how to add full support for `resolutions`, potentially using functionality provided by `@yarnpkg/core`.

[...]

PR #102, opened 2021-04-19
fix: allow yarn 2 resolutions when strict mode is disabled

JamesPatrickGill:
- [ ] Reviewed by Snyk team

#### What this does
Yarn 2 resolutions breaks snyk test. See #99 for details. This PR unblocks consumers that have yarn 2 resolutions when --strict-out-of-sync=false is passed to the Snyk CLI.

Notes for the reviewer
See additional test case for yarn 1 and yarn 2` resolutions.


test/lib/fixtures/resolutions/yarn2/expected-tree.json
```diff
@@ -0,0 +1,20 @@
+{
+  "dependencies": {
+    "adm-zip": {
+      "labels": {
+        "scope": "prod",
+        "missingLockFileEntry": "true"
+      },
+      "name": "adm-zip",
+      "version": "0.4.8"
```
JamesPatrickGill, line 9: I have changed the tests on this line to reflect expected behaviour - that the package version in `resolutions` is what snyk picks up. 

This is because the resolution package will be the version the user will be using within their code base so makes most sense to make sure we scan it.

mhassan1: I agree that is the correct behavior, but it breaks the test. I wasn't attempting to fully solve the `resolutions` use case in the original PR. I was just hoping to unblock Yarn 2 users.

JamesPatrickGill: Hey @mhassan1, sorry for not being clearer, we tested your proposal against this test case to highlight what we expected based on what we understood from the docs ([resolutions](https://classic.yarnpkg.com/en/docs/package-json/#toc-resolutions)) and the behaviour of snyk. 

We expected that if we scan a project with resolutions snyk will scan the resolutions versions for vulnerabilities , but we see that it scans the version in the dependencies instead which could lead to missed vulnerabilities which is a problem for us.

We don’t have capacity to figure this out this sprint, so if you can come up with a solution sooner than we can get to this, we would highly appreciate it. Moreover, if our understanding is wrong please feel free to correct us.

mhassan1: I understand that missed vulnerabilities due to `resolutions` is a problem for Snyk, but this PR was not meant to fix that issue. The `resolutions` problem has always been unhandled with both Yarn 1 and Yarn 2 in the past. This PR was only to unblock Yarn 2 users (who right now can't use the latest version of the Snyk CLI at all if they have any `resolutions`). I think a working Snyk CLI that doesn't correctly consider `resolutions` is better than a broken Snyk CLI, but feel free to disagree.

I'm happy to help figure out the `resolutions` problem, but that may be a harder problem to get right, since the logic for `resolutions` may be complicated. I think that conversation is better suited to a new PR.

JamesPatrickGill:
Hi, full support for resolutions in yarn2 has been introduced in v1.35 of this parser via [this PR](https://github.com/snyk/nodejs-lockfile-parser/pull/115). 

It is included in version 1.616.0 of the CLI 🎉 

Closing this PR as a result, thank you for your contribution and insight into this problem 😃

PR #110, opened 2021-05-12
feat: correct dependencies versions are scanned from resolutions

JamesPatrickGill:
- [ ] Reviewed by Snyk team

### What this does

This add support for the use of resolutions in the `package.json`, IS WIP


test/fixtures/resolutions/yarn2/yarn.lock
```diff
@@ -0,0 +1,21 @@
+# This file is generated by running "yarn install" inside your project.
```
admons, line 1: where's the `package.json` file in this folder?

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -41,8 +44,18 @@ export class Yarn2LockParser extends LockParserBase {
 
       Object.entries(rawYarnLock).forEach(
         ([fullDescriptor, versionData]: [string, any]) => {
+          const newVersionData =
+            manifestFile.resolutions && versionData.dependencies
```
admons, line 48: Doesn't this means that we'll add all **root manifest file**'s resolutions to each dependency **as a dependency**?

Please add more complex fixtures and I think you'll see the error

admons (changes requested):
In addition to the comments I added, `resolutions` has a way to apply globs to specify exactly which packages will be affected: https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md#package-designation

test/fixtures/resolutions/yarn1/expected-tree.json
```diff
@@ -0,0 +1,19 @@
+{
+  "dependencies": {
+    "adm-zip": {
+      "labels": {
+        "scope": "prod"
+      },
+      "name": "adm-zip",
+      "version": "0.4.8"
```
admons, line 8: According to [this](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md#relation-to-non-nested-dependencies), direct dependencies (dev, regular, optional) will always take precedence over the `resolutions`

PR #115, opened 2021-05-21
feat: add support for resolutions in yarn2

JamesPatrickGill:
- [ ] Reviewed by Snyk team

## What this does

This allows Snyk to scan yarn2 projects that use resolutions.

## Note

Resolutions behave interestingly and there is not much documentation available for how yarn2 resolutions work (all i could find was [this section](https://yarnpkg.com/configuration/manifest/#resolutions) of a bigger doc). As such the reasoning of the code may seem weird at first - to help this also want to outline keypoints of how resolutions work in yarn2.

### Glob patterns in resolutions are gone:

Resolutions now only allow one level of specificity. This means you can either just use a package name, or a package and its parent to specify a resolution as such 

```javascript
{
    "debug": "4.0.0", // good
    "express/debug": "4.0.0", // good
    "express/**/debug": "4.0.0" // bad
}
```

This means this ticket does not need to worry about globs.

### Order matters (_...really_):

Inside the resolutions object the type of key used to specify a resolution does not give it priority over another if there are multiple matches, it is whichever is first, i.e:

```javascript
{
    "debug": "4.0.0", 
    "express/debug": "3.0.0",
}
```

will result in only `debug@4.0.0` being used whereas:

```javascript
{
    "express/debug": "3.0.0",
    "debug": "4.0.0", 
}
```

will result in both `debug@4.0.0` *and*  `debug@3.0.0` as 3.0.0 will be used for direct dependent of express only.

### Qualifiers can be used in resolutions

This is where you can specify to only apply a resolution if the parent of the resolution is being used at a certain version. This looks like this:

```javascript
{
   "@babel/core@npm:7.0.0/@babel/generator": "7.3.4",
}
```

This is why this weird check exists: https://github.com/snyk/nodejs-lockfile-parser/blob/feat/resolutions-for-yarn2/lib/parsers/yarn2-lock-parser.ts#L128-L131

Please let me know if you would like to go through this with me.


lib/parsers/yarn2-lock-parser.ts
```diff
@@ -81,21 +93,57 @@ export class Yarn2LockParser extends LockParserBase {
...
+    resolutions?: ManifestDependencies,
+  ): DepMap {
     const yarnLockfile = lockfile as Yarn2Lock;
     const depMap: DepMap = {};
 
+    const dependencies = lockfile.dependencies || {};
+    const resolutionPaths = resolutions ? Object.keys(resolutions) : [];
+
     for (const [depName, dep] of Object.entries(yarnLockfile.object)) {
+      const depNameNoVer = depName
```
admons, line 107: You can find a better way to do this :)
The only time you'll have more than one @ character is when the package is scoped (I.e. @snyk/something)
You can easily solve it with regex, something like '/.*@.*/' should work

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -81,21 +93,57 @@ export class Yarn2LockParser extends LockParserBase {
...
+        .join('@');
+
+      const depNameVersion = depName.split('@')[depName.split('@').length - 1];
+
       const subDependencies = Object.entries({
         ...(dep.dependencies || {}),
         ...(dep.optionalDependencies || {}),
+      }).map(([key, ver]) => {
+        const possibleResolutions =
+          resolutionPaths.length &&
```
admons, line 119: No need to check this

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -81,21 +93,57 @@ export class Yarn2LockParser extends LockParserBase {
...
+          resolutionPaths.filter((res) => {
```
admons, line 120: You can use .find instead of filter

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -81,21 +93,57 @@ export class Yarn2LockParser extends LockParserBase {
...
 
-  protected getDepMap(lockfile: Lockfile): DepMap {
+  protected getDepMap(
+    lockfile: Lockfile,
+    resolutions?: ManifestDependencies,
+  ): DepMap {
     const yarnLockfile = lockfile as Yarn2Lock;
     const depMap: DepMap = {};
 
+    const dependencies = lockfile.dependencies || {};
```
admons, line 103: Where's the logic for * and **?

admons (approved):

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -81,21 +90,36 @@ export class Yarn2LockParser extends LockParserBase {
...
+
     for (const [depName, dep] of Object.entries(yarnLockfile.object)) {
       const subDependencies = Object.entries({
         ...(dep.dependencies || {}),
         ...(dep.optionalDependencies || {}),
+      }).map(([key, ver]) => {
```
admons, line 106: usually for one liners we're using 
``` (args...) => result```
instead of
``` 
(...args) => {
  return result;
}
```

File: lib/parsers/yarn2-lock-parser.ts [hunk 6]
```diff
@@ -111,3 +135,44 @@ export class Yarn2LockParser extends LockParserBase {
...
+
+function findResolutions(
+  dependencies: YarnLockDeps,
+  depName: string,
+  subDepKey: string,
+  resolutions?: ManifestDependencies,
+): string | undefined {
+  if (!resolutions) return;
+
+  const resolutionKeys = resolutions ? Object.keys(resolutions) : [];
```
admons, line 147: No need to recheck resolutions

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -111,3 +135,44 @@ export class Yarn2LockParser extends LockParserBase {
...
+      specifiedParentMatchesCurrentDep ||
+      specifiedParentWithVersionMatches
+    );
+  });
+
+  if (resolutionKeys && firstMatchingResolution) {
+    return `${subDepKey}@${resolutions[firstMatchingResolution]}`;
+  }
+
+  return;
```
admons, line 177: This is redundant

lib/parsers/yarn2-lock-parser.ts
```diff
@@ -81,21 +90,36 @@ export class Yarn2LockParser extends LockParserBase {
...
 
-  protected getDepMap(lockfile: Lockfile): DepMap {
+  protected getDepMap(
+    lockfile: Lockfile,
+    resolutions?: ManifestDependencies,
+  ): DepMap {
     const yarnLockfile = lockfile as Yarn2Lock;
     const depMap: DepMap = {};
 
+    const dependencies = lockfile.dependencies || {};
```
admons, line 100: Add `YarnLockDeps` type (instead of casting later)

'
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env node_modules/.bin/ts-node
```
admons, line 1: 1. all other files are not using this shebang, I think it's better to keep it consistent
2. I think you can find a better name for this file 😂

lib/parsers/yarn2-lock-parser.ts
JamesPatrickGill: Cast is still necessary as a type annotation on variable to powerful to cast from this type for whatever reason: `dependencies: PackageLockDeps | YarnLockDeps`

JamesPatrickGill: Moved cast to this line though to make it more sensible

PR #118, opened 2021-08-10
fix: pass includePeerDeps instead of a whole lockfile

lili2311:
- [x] Reviewed by Snyk team

### What this does

Instead of passing an entire lockfile to a function that only uses it
to decide if we need to grab peer deps, do this calculation upfront
and optionally allow grabbing peer deps. This also allows for future support
of --peer

### DIFF
--- a/lib/parsers/index.ts
+++ b/lib/parsers/index.ts
@@ -102,7 +102,7 @@ export function parseManifestFile(manifestFileContents: string): ManifestFile {
 export function getTopLevelDeps(
   targetFile: ManifestFile,
   includeDev: boolean,
-  lockfile: Lockfile,
+  includePeerDeps = false,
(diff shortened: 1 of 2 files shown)


lib/parsers/index.ts
```diff
@@ -102,7 +102,7 @@ export function parseManifestFile(manifestFileContents: string): ManifestFile {
 export function getTopLevelDeps(
   targetFile: ManifestFile,
   includeDev: boolean,
-  lockfile: Lockfile,
+  includePeer = false,
```
jan-stehlik, line 105:
```suggestion
  includePeerDeps = false,
```

jan-stehlik (approved):
nice one @lili2311 , thanks 💪 🎉

PR #126, opened 2021-09-28
fix: applies resolutions when finding top level deps if yarn 2

JamesPatrickGill:
- [ ] Reviewed by Snyk team

### What this does

This makes a small fix to apply resolutions when finding top level deps. Currently scoped just to yarn2 for first rollout.

### DIFF
--- a/lib/parsers/index.ts
+++ b/lib/parsers/index.ts
@@ -99,12 +99,18 @@ export function parseManifestFile(manifestFileContents: string): ManifestFile {
   }
 }
 
-export function getTopLevelDeps(
-  targetFile: ManifestFile,
-  includeDev: boolean,
+export function getTopLevelDeps({
+  targetFile,
+  includeDev,
   includePeerDeps = false,
-): Dep[] {
-  const dependencies: Dep[] = [];
+  applyYarn2Resolutions = false,
+}: {
+  targetFile: ManifestFile;
+  includeDev: boolean;
+  includePeerDeps?: boolean;
+  applyYarn2Resolutions?: boolean;
+}): Dep[] {
+  let dependencies: Dep[] = [];
 
   const dependenciesIterator = Object.entries({
     ...targetFile.dependencies,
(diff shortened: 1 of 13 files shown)


lib/parsers/index.ts
```diff
@@ -131,6 +132,22 @@ export function getTopLevelDeps(
       });
     }
   }
+
+  if (applyResolutions && targetFile.resolutions) {
```
jan-stehlik, line 136: can we please add a test for this?

JamesPatrickGill: Should be tested now

jan-stehlik (approved):
nice one @JamesPatrickGill :rocket:  i have only a suggestion

lib/parsers/lock-parser-base.ts
```diff
@@ -137,6 +137,7 @@ export abstract class LockParserBase implements LockfileParser {
       manifestFile,
       includeDev,
       lockfile.type === LockfileType.npm7,
+      lockfile.type === LockfileType.yarn2,
```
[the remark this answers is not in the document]
JamesPatrickGill: I find this much more readable but as this is an exported function, it changes this libs api and was trying to make as little impact as possible,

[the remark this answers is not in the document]
jan-stehlik: it's not exported as a public API for consumers of the library though https://github.com/snyk/nodejs-lockfile-parser/blob/009ac12a356e4d005ae40b2956feef4dbf1d9a3d/lib/index.ts#L23-L35 right? change is completely up to you though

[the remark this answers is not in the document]
JamesPatrickGill: Ahh i miss remembered, my bad

lib/parsers/index.ts
```diff
@@ -131,6 +132,22 @@ export function getTopLevelDeps(
...
+    for (const [resName, resVersion] of Object.entries(
```
admons, line 137: What do you think about this approach?

```typescript
const map = new Map(Object.entries(targetFile.resolutions).map(
    ([resName, resVersion]) => [resName.replace(`${targetFile.name}/`, ''), resVersion]))

return dependencies.map(dep =>  map.has(dep.name) ? {...dep, version: map.get(dep.name)} : dep)
```

JamesPatrickGill: followed this approach

PR #141, opened 2022-06-02
fix: display unknown-version instead of truncating package name when …

xzhou-snyk:

### What this does

For the following npm lock file where packages exist in `requires` list but miss the lock entries, the package resolution displays weirdly:

```
"parentPackage": {
  "version": "0.0.1",
    "requires": {
      "childPackage1": "^1.0.0",
      "childPackage2": "^1.0.0"
    }
}

// no package resolution for childPackage1, childPackage2
```

It displays as `childPackage@childPackage1` and `childPackage@childPackage2`. This is due to that we take away the version info (the `"^1.0.0"` part) from the package in `requires` list when constructing the dep map, which makes sense because in theory there should be corresponding lock entries in the lock file for these packages which would contain the actual resolved version. Then `indexOf('@')` for the `childPackage1` resolves to `-1`.

This PR slightly improves the display by showing `unknown-version`, i.e. `childPackage1@unknown-version` and `childPackage2@unknown-version`.

This PR also uses the [npm-package-arg](https://github.com/npm/npm-package-arg) to parse the package name instead of relying on `indexOf('@')`.

### Screenshots
Screenshot 2022-06-02 at 17 56 57

### DIFF
--- a/lib/cli-parsers/cli-parser-utils.ts
+++ b/lib/cli-parsers/cli-parser-utils.ts
@@ -1,12 +1,11 @@
 import * as semver from 'semver';
+import * as npa from 'npm-package-arg';
 
 export const extractNameAndIdentifier = (
   candidate: string,
 ): { name: string; identifier: string } => {
-  const index = candidate.indexOf('@', 1);
-  const name = candidate.slice(0, index);
-  const identifier = candidate.slice(index + 1);
-  return { name, identifier };
+  const parsed = npa(candidate);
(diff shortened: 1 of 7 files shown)


JamesPatrickGill:
Hi, wondering about the scenario that would lead to this lockfile being used, is this even a valid lockfile? I ask as I don't think we should be glossing over a broken lockfile and making it "work". IMO we should error out with invalid lock.

tonidevine1 (approved):
Great work!

PR #146, opened 2022-06-08
feat: remove npa package name parser and revert to old method

xzhou-snyk:
This PR removes the `npm-package-arg` package and reverts to the old method when parsing package name. The reason is, `npm-package-arg` is excellent in parsing package names in `package-lock.json`, but it couldn't parse certain complicated package names in `yarn2.lock`, e.g. `resolve@patch:resolve@^1.10.0#builtin<compat/resolve>` which is using the compat plugin.


tonidevine1 (approved):
LGTM

PR #150, opened 2022-08-06
Feat/use dfs to detect cycle and build graph

xzhou-snyk:
This PR uses DFS to traverse all package nodes and detect cyclic dependencies.

The reason for the improvement is, the `ancestorMap` in the original solution could be expensive as it will copy ancestorMap entries a lot of times. Also there is actually standard solution for detecting cycles using DFS.

This PR also did some refactoring and grouped some reused functions together :)


lib/dep-graph-builders/util.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const devDeps = options.includeDevDeps
+    ? getGraphDependencies(pkgJson.devDependencies || {}, true)
+    : {};
+
+  return { ...prodDeps, ...devDeps };
+};
+
+/**
+ * Converts dependencies parsed from the yarn.lock file to a dependencies object required by the graph.
```
JamesPatrickGill, line 47: yarn.lock is incorrect here right? Otherwise this isn't a general function

lib/dep-graph-builders/util.ts
```diff
@@ -0,0 +1,64 @@
...
+  name: string;
+  version: string;
+  dependencies: Record<string, { version: string; isDev: boolean }>;
+  isDev: boolean;
+}
+
+export const addPkgNodeToGraph = (
+  node: PkgNode,
+  isCyclic: boolean,
+  depGraphBuilder: DepGraphBuilder,
```
JamesPatrickGill, line 15: Can we put the mutated element as first argument, keeps it clear in future if we add any more. Or return the depgraphbuilder to keep it explicit.

lib/dep-graph-builders/util.ts
xzhou-snyk: this is so well spotted :))) thank you :)))

lib/dep-graph-builders/util.ts
xzhou-snyk: good point !! also will change argument order for the other methods where `depGraphBuilder` is used ;)

JamesPatrickGill: Thank you, just mutating args is a bit hard to see straight away so lets make it as clear as possible

PR #166, opened 2022-11-29
Feat/yarn lock v2 support straight to dg

JamesPatrickGill:


lib/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+export function getNpmLockfileVersion(
+  lockFileContents: string,
+):
+  | NodeLockfileVersion.NpmLockV1
+  | NodeLockfileVersion.NpmLockV2
+  | NodeLockfileVersion.NpmLockV3 {
+  if (lockFileContents.includes(`"lockfileVersion": 1,`)) {
```
admons, line 44: looking at the history of the `package-lock.json` file, is it possible that `lockfileVersion` didn't exists and we should consider it as v1?

admons (approved):

lib/utils.ts
JamesPatrickGill: Reasonable

PR #185, opened 2023-03-28
fix: correctly index pkg lock with pkgs with similar candidates

JamesPatrickGill:
This addition, means that we further fix indexing into the pkg lock `packages` option when a dependency has a number of similar candidates.


yurinka1:
@JamesPatrickGill Do you mind adding tests to cover the use cases?

And an example for noobies like me 🙏🏻

yurinka1 (approved):

PR #187, opened 2023-03-31
fix: accuracy improvements for npm lock v2 and v3 parsing

JamesPatrickGill:
This improves upon the resolution when parsing npm lock v2 and v3 files. It also adds comments to the function in question to help illustrate the goals.


lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -235,85 +242,79 @@ const getChildNode = (
...
-        key.includes(parentKey),
-      );
+    // So now we can check semver to filter out some values
+    const candidatePkgVersion = pkgs[candidate].version;
+    keep = semver.satisfies(candidatePkgVersion, version);
 
-      if (possibleFilteredKeys.length === 1) {
-        return possibleFilteredKeys[0];
-      }
+    if (keep === false) {
```
admons, line 306: you can skip this test as you're returning keep right after

test/jest/dep-graph-builders/fixtures/npm-lock-v2/deeply-nested-packages/expected.json
```diff
@@ -3892,7 +3906,7 @@
         "pkgId": "string_decoder@1.1.1",
         "deps": [
           {
-            "nodeId": "safe-buffer@5.2.0"
+            "nodeId": "safe-buffer@5.1.2"
```
admons, line 3909: Nice!

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -235,85 +242,79 @@ const getChildNode = (
...
+    ...ancestry.slice(rootOperatingIdx).map((el) => el.name),
+    name,
+  ];
+
+  // We filter on a number of cases
+  let filteredCandidates = candidateKeys.filter((candidate) => {
+    let keep = true;
+
+    const candidateAncestry = candidate
+      .replace('node_modules/', '')
```
admons, line 278: Are you sure about this one? I don't know how's the candidate string looks like, but this is confusing. maybe add a comment to explain?

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -235,85 +242,79 @@ const getChildNode = (
...
+    ? ancestry.findIndex((el) => el.inBundle === true) - 1
+    : 1;
+  const ancestryFromRootOperatingIdx = [
+    ...ancestry.slice(rootOperatingIdx).map((el) => el.name),
+    name,
+  ];
+
+  // We filter on a number of cases
+  let filteredCandidates = candidateKeys.filter((candidate) => {
+    let keep = true;
```
admons, line 275: Would it make more sense to use ad-hoc variables that explains why you want to bail out instead of having one "keep" "global" variable?

PR #191, opened 2023-05-18
feat: add pruning to yarn lock v2 dep graph builder

JamesPatrickGill:

This adds pruning within top level dependencies to the Yarn Lockfile Version 2 parser


lib/dep-graph-builders/yarn-lock-v2/build-depgraph-simple.ts
```diff
@@ -76,19 +80,37 @@ const dfsVisit = async (
...
-        resolutions,
-      );
+    if (localVisited.has(childNode.id)) {
+      if (pruneWithinTopLevel) {
+        const prunedId = `${childNode.id}:pruned`;
+        depGraphBuilder.addPkgNode(
+          { name: childNode.name, version: childNode.version },
+          prunedId,
+          { labels: { scope: node.isDev ? 'dev' : 'prod', pruned: 'true' } },
+        );
```
gitphill, line 90: what was wrong with `addPkgNodeToGraph`?

test/jest/dep-graph-builders/yarn-lock-v2.test.ts
```diff
@@ -1,58 +1,444 @@
...
+            `./fixtures/yarn-lock-v2/real/${fixtureName}/expected.json`,
+          ),
+          'utf8',
+        ),
+      );
+
+      expect(dg).toBeTruthy();
+
+      const expectedDepGraph = createFromJSON(expectedDepGraphJson);
+      expect(dg.equals(expectedDepGraph)).toBeTruthy();
```
gitphill, line 55: test fixtures didn't change! does that mean non of these real test fixtures had pruning opportunities, and/or does this equals not care about prune labels

JamesPatrickGill: Must be the former

gitphill: i can't believe goof cannot be pruned

lib/dep-graph-builders/yarn-lock-v2/build-depgraph-simple.ts
JamesPatrickGill: I don't think the abstraction is providing value, and wasn't built with nodeId and pkgId being different in mind

JamesPatrickGill: I can bring it back if need be

test/jest/dep-graph-builders/yarn-lock-v2.test.ts
JamesPatrickGill: The old tests have pruning off in the options, it's a flag you pass in this eco

JamesPatrickGill: https://github.com/snyk/nodejs-lockfile-parser/pull/191/files#diff-c09b371ecae0c6d79c228da21ef8f9ae6932a8ccd45472b23a1b60b8799eaae7R29-R34

test/jest/dep-graph-builders/yarn-lock-v2.test.ts
```diff
@@ -1,58 +1,444 @@
...
+        nodeId: 'b@1.0.0',
+        pkgId: 'b@1.0.0',
+        deps: [{ nodeId: 'a@1.0.0:pruned' }],
+        info: { labels: { scope: 'prod' } },
+      },
+      {
+        nodeId: 'a@1.0.0:pruned',
+        pkgId: 'a@1.0.0',
+        deps: [],
+        info: { labels: { scope: 'prod', pruned: 'true' } },
```
gitphill, line 250: is it pruned: true or cyclic https://github.com/snyk/dep-graph/blob/master/src/core/types.ts#L51 do we not really care?

JamesPatrickGill: I no longer care

test/jest/dep-graph-builders/yarn-lock-v2.test.ts
gitphill: are we ever going to not prune? whats the value in those tests

JamesPatrickGill: I believe we should test both, this library may be used in applications that don not wish to prune

gitphill (approved):

dotkas (approved):
image

PR #192, opened 2023-05-19
feat: add pruning to yarn lock v1 dep graph builder

JamesPatrickGill:

### What this does

Adds pruning options to the yarn lock v1 dep graph builder


lib/dep-graph-builders/yarn-lock-v1/build-depgraph-simple.ts
```diff
@@ -54,17 +57,17 @@ export const buildDepGraphYarnLockV1Simple = async (
...
+  pruneWithinTopLevel: boolean,
+  visited?: Set<string>,
 ): Promise<void> => {
-  visitedMap.add(node.id);
-
   for (const [name, depInfo] of Object.entries(node.dependencies || {})) {
     if (eventLoopSpinner.isStarving()) {
       await eventLoopSpinner.spin();
     }
+    const localVisited = visited || new Set<string>();
```
dotkas, line 70: nit, but you could consider moving this out of the loop to avoid constant object creation. Not sure how bit a hit it has on the performance though.

lib/dep-graph-builders/yarn-lock-v1/build-depgraph-simple.ts
```diff
@@ -73,18 +76,51 @@ const dfsVisit = async (
...
 
+    depGraphBuilder.addPkgNode(
+      { name: childNode.name, version: childNode.version },
+      childNode.id,
+      {
+        labels: {
+          scope: node.isDev ? 'dev' : 'prod',
+          ...(node.missingLockFileEntry && {
+            missingLockFileEntry: 'true',
+          }),
```
dotkas, line 110: Same with this. Though very nitpick, so up to you.

dotkas (approved):
You the real MVP. Most Valuable Pruner.

lib/dep-graph-builders/yarn-lock-v1/build-depgraph-simple.ts
JamesPatrickGill: The reason it is in the loop is to create it on each top level iteration, after first recursion layer though it will use the `visited` passed in as a function arg to `dfsVisit`

lib/dep-graph-builders/yarn-lock-v1/build-depgraph-simple.ts
JamesPatrickGill: I don't know an alternative to this being here it needs to be called every time the code gets to this point. If you are talking repeat adding, adding a repeat node is as close to a noop as we get in js land

PR #196, opened 2023-06-28
fix: correct bundled deps behaviour and add fixture

JamesPatrickGill:
### What this does

Fixes bundled deps resolution mistake where full filepath (`key`) was used instead of the name


test/jest/dep-graph-builders/npm-lock-v2.test.ts
```diff
@@ -338,3 +338,32 @@ describe('dep-graph-builder npm-lock-v2', () => {
...
+      'utf8',
+    );
+    const npmLockContent = readFileSync(
+      join(
+        __dirname,
+        `./fixtures/npm-lock-v2/${fixtureName}/package-lock.json`,
+      ),
+      'utf8',
+    );
+    // this throws OutOfSyncError! why...
```
gitphill, line 356: Can remove this

test/jest/dep-graph-builders/npm-lock-v2.test.ts
```diff
@@ -338,3 +338,32 @@ describe('dep-graph-builder npm-lock-v2', () => {
...
+      pkgJsonContent,
+      npmLockContent,
+      {
+        includeDevDeps: false,
+        includeOptionalDeps: true,
+        pruneCycles: true,
+        strictOutOfSync: true,
+      },
+    );
+    expect(depGraph).toBeTruthy();
```
gitphill, line 367: Is there a better assert

JamesPatrickGill: Better one added

orsagie (approved):

PR #202, opened 2023-09-07
fix: resolve nested deps to matching version

klesniewski:
- [X] Tests written and linted
- [X] Follows [CONTRIBUTING agreement](CONTRIBUTING.md)
- [X] Commit history is tidy [https://git-scm.com/book/en/v2/Git-Branching-Rebasing](i)
- [X] Reviewed by Snyk team

### What this does

* Fixes resolving dependencies in NPM lockfile v2. In some cases, when multiple versions of a dependency are present, in some places a wrong version was picked up.
* Adds testing against `npm ls` output, so we know our results are consistent.
* Fixes test fixtures to match `npm ls` output.


test/jest/dep-graph-builders/npm-lock-v2.test.ts
```diff
@@ -40,14 +46,36 @@ describe('dep-graph-builder npm-lock-v2', () => {
             },
           );
 
+          if (existsFixtureFileSync(fixturePath, 'expected-npm-ls.txt')) {
```
klesniewski, line 49: If `expected-npm-ls` file is provided, then we evaluate also against that.

PR #207, opened 2023-11-07
fix: correctly pick npm-lock-v2 pkg version and prune within subtrees only

JamesPatrickGill:
### What this does

There are two fixes in this PR:

1. Fixed a bug in the choosing of a child node key which did not respect the already determined possible keys.
2. Scoped pruning to the subtrees of top level dependencies so that we don't lose context.

### DIFF
--- a/lib/dep-graph-builders/npm-lock-v2/index.ts
+++ b/lib/dep-graph-builders/npm-lock-v2/index.ts
@@ -7,7 +7,6 @@ import { extractPkgsFromNpmLockV2 } from './extract-npm-lock-v2-pkgs';
 import type { NpmLockPkg } from './extract-npm-lock-v2-pkgs';
 import { DepGraph, DepGraphBuilder } from '@snyk/dep-graph';
 import {
-  addPkgNodeToGraph,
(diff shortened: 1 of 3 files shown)


DOlufemi (approved):

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -88,35 +89,35 @@ export const buildDepGraphNpmLockV2 = async (
...
     npmLockPkgs,
     strictOutOfSync,
     includeDevDeps,
     includeOptionalDeps,
     [],
     pkgKeysByName,
   );
-  return depGraphBuilder.build();
+  const dg = depGraphBuilder.build();
+  return dg;
```
yurinka1, line 103: Why?

JamesPatrickGill: Who knows, I will change

JamesPatrickGill:
Closing as separated fixes out into their own PRs https://github.com/snyk/nodejs-lockfile-parser/pull/208

PR #212, opened 2023-11-23
fix: start supporting overrides in a non destructive way - npm lock v2

JamesPatrickGill:

### What this does

Newer changes to npm-lock-v2 resolution started to fail in cases where overrides where used. Although we never had explicit support for this, the old behavior was at least not a failure. 

To remediate I have just started to explicitly support these use cases, there are likely rough edges but for now this should help support some existing use cases I have found.


lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -379,3 +401,34 @@ export const getChildNodeKey = (
...
+const checkOverrides = (
+  ancestry: Ancestry[],
+  overrides: Overrides,
+): string | undefined => {
+  const ancestryWithoutRoot = ancestry.slice(1);
+  // First traverse into overrides from root down
+  for (const [idx, pkg] of ancestryWithoutRoot.entries()) {
+    // Do we have this in overrides
+    const value =
+      overrides[pkg.name] || overrides[`${pkg.name}@${pkg.version}`];
```
dotkas, line 413:
```suggestion
    // Do we have this in overrides
    const key = pkg.name] || overrides[`${pkg.name}@${pkg.version}`;
    const value =
      overrides[key];
```

nitpicking

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -379,3 +401,34 @@ export const getChildNodeKey = (
...
+
+    if (value) {
```
dotkas, line 415: Reverse logic and return early

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -379,3 +401,34 @@ export const getChildNodeKey = (
...
+      if (typeof value !== 'string') {
+        // Check for dotted resolution
+        if (idx + 1 === ancestryWithoutRoot.length && value['.']) {
+          return value['.'];
+        }
```
dotkas, line 420: Needs explanation

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -379,3 +401,34 @@ export const getChildNodeKey = (
...
+
+        const res = checkOverrides(ancestryWithoutRoot, value);
+        if (res) {
+          return res;
+        }
+        continue;
+      } else {
+        return value;
+      }
```
dotkas, line 429: Reverse logic and return early

```suggestion
        if (!res) {
          continue;
        }
        return res;
      }
      return value;
      }
```

dotkas (approved):
Nitpicked your code style as I had nothing of substance to add.

test/jest/dep-graph-builders/npm-lock-v2.test.ts
```diff
@@ -56,6 +56,55 @@ describe('dep-graph-builder npm-lock-v2', () => {
...
+              readFileSync(
+                join(
+                  __dirname,
+                  `./fixtures/npm-lock-v2/${fixtureName}/expected.json`,
+                ),
+                'utf8',
+              ),
+            );
+            const expectedDepGraph = createFromJSON(expectedDepGraphJson);
+            expect(newDepGraph.equals(expectedDepGraph)).toBeTruthy();
```
dotkas, line 103: Just because the `diff` is easier to view when this fails, I usually do

```suggestion
            expect(newDepGraph.toJSON().toEqual(expectedDepGraph.toJSON());
```

lib/dep-graph-builders/npm-lock-v2/index.ts
JamesPatrickGill: Resolutions can be written as a dot inside the object with its key as name.

To resolve this I have to know 
1. Are we looking at the package that we actually want to find a resolution for - this is the index check, we need to know we are at the last index of the ancestry array.
2. I then need to make sure the ['.'] value even exists.

test/jest/dep-graph-builders/npm-lock-v2.test.ts
JamesPatrickGill: This doesn't account for changes in shape but otherwise equivalent depgraphs though. Think shape isn't the only thing that explains equivalence.

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -379,3 +401,34 @@ export const getChildNodeKey = (
...
+  const ancestryWithoutRoot = ancestry.slice(1);
+  // First traverse into overrides from root down
+  for (const [idx, pkg] of ancestryWithoutRoot.entries()) {
+    // Do we have this in overrides
+    const value =
+      overrides[pkg.name] || overrides[`${pkg.name}@${pkg.version}`];
+
+    if (value) {
+      if (typeof value !== 'string') {
+        // Check for dotted resolution
```
dotkas, line 417:
```suggestion
// Resolutions can be written as a dot inside the object with its key as name.

// To resolve this I have to know

// Are we looking at the package that we actually want to find a resolution for - this is the index check, we need to know we are at the last index of the ancestry array.
I then need to make sure the ['.'] value even exists.
```

PR #213, opened 2023-11-28
fix: more accurately resolves deps of overrides

JamesPatrickGill:
### What this does

This fixes dependencies of resolutions being given the override version of their parent.

### DIFF
--- a/lib/dep-graph-builders/npm-lock-v2/index.ts
+++ b/lib/dep-graph-builders/npm-lock-v2/index.ts
@@ -418,8 +418,12 @@ const checkOverrides = (
       continue;
     }
 
-    // If we find a string as override we know we found what we want
-    if (typeof override === 'string') {
+    // If we find a string as override we know we found what we want *if*
+    // we are at our root
+    if (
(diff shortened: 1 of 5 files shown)


lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -418,8 +418,12 @@ const checkOverrides = (
...
     }
 
-    // If we find a string as override we know we found what we want
-    if (typeof override === 'string') {
+    // If we find a string as override we know we found what we want *if*
+    // we are at our root
+    if (
+      idx + 1 === ancestryWithoutRoot.length &&
+      typeof override === 'string'
+    ) {
```
darscan, line 426: tests?

JamesPatrickGill: Looking to create some fixtures now to illustrate this behaviour

JamesPatrickGill: Done

darscan (approved):

PR #214, opened 2023-12-07
fix: allow overrides with version specs

JamesPatrickGill:
### What this does

Please look at test fixtures to understand goal.

This allows us to work with simple overrides which have a version specifier in the name.

### DIFF
--- a/lib/dep-graph-builders/npm-lock-v2/index.ts
+++ b/lib/dep-graph-builders/npm-lock-v2/index.ts
@@ -410,8 +410,7 @@ const checkOverrides = (
   // First traverse into overrides from root down
   for (const [idx, pkg] of ancestryWithoutRoot.entries()) {
     // Do we have this in overrides
-    const override =
-      overrides[pkg.name] || overrides[`${pkg.name}@${pkg.version}`];
+    const override = matchOverrideKey(overrides, pkg);
 
     // If we dont find current element move down the ancestry
     if (!override) {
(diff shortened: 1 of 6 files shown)


lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -444,3 +443,42 @@ const checkOverrides = (
   }
   return;
 };
+
+// Here we have to match our pkg to
+// possible keys in the overrides object
+const matchOverrideKey = (
```
anthogez, line 449: tests for matchOverrideKey?

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -444,3 +443,42 @@ const checkOverrides = (
...
+  overrides: Overrides,
+  pkg: { name: string; version: string },
+): string | null => {
+  if (overrides[pkg.name]) {
+    return overrides[pkg.name];
+  }
+
+  const overrideKeysNameToVersions = Object.keys(overrides).reduce(
+    (acc, key) => {
+      const splitNameAndVersion = key.split('@');
```
anthogez, line 459: will always be the case where split per `@` will be valid?

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -444,3 +443,42 @@ const checkOverrides = (
...
   return;
 };
+
+// Here we have to match our pkg to
+// possible keys in the overrides object
+const matchOverrideKey = (
+  overrides: Overrides,
+  pkg: { name: string; version: string },
+): string | null => {
+  if (overrides[pkg.name]) {
```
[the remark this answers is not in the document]
JamesPatrickGill: If theres one with just name it wins

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -444,3 +443,42 @@ const checkOverrides = (
...
+  const overrideKeysNameToVersions = Object.keys(overrides).reduce(
+    (acc, key) => {
+      const splitNameAndVersion = key.split('@');
+      const versionSpec = splitNameAndVersion.pop();
+      const name = splitNameAndVersion.join('@');
+      const versionsArr = acc[name]
+        ? [...acc[name], versionSpec]
+        : [versionSpec];
+      return { ...acc, [name]: versionsArr };
+    },
```
anthogez, line 466: How about?

```
  // Split the key to separate the package name from the version spec
  const atIndex = key.lastIndexOf('@');
  const name = key.substring(0, atIndex);
  const versionSpec = key.substring(atIndex + 1);

  // Check if the package name already exists in the accumulator
  if (!acc[name]) {
    acc[name] = [];
  }

  // Add the version spec to the list of versions for this package name
  acc[name].push(versionSpec);

  return acc;
```

lib/dep-graph-builders/npm-lock-v2/index.ts
JamesPatrickGill: Can write a specific unit test

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -444,3 +443,42 @@ const checkOverrides = (
...
+    {},
+  );
+
+  if (overrideKeysNameToVersions[pkg.name]) {
+    for (const versionSpec of overrideKeysNameToVersions[pkg.name]) {
```
anthogez, line 471: why are we increasing computation time twice?

```
const keysNameToVersion = overrideKeysNameToVersions[pkg.name];

then reuse keysNameToVersion both for line 470 and line 471?
```

JamesPatrickGill: Fair enough, was an extra hash lookup so didn't delve in too much

anthogez (approved):

PR #215, opened 2023-12-08
fix: resolve npm scoped versions

JamesPatrickGill:
### What this does

Previously if this parser saw a version specifier like `npm:something@^1.0.0` this whole specifier was used as the version in the depgraph. This is a bit disingenuous as we should be able to resolve the version this actually gets resolved to using the lockfile. This behavior also causes downstream issues in the Snyk CLI.

Now these versions will attempt to be resolved to just the final version range in the specifier.

### DIFF
--- a/lib/dep-graph-builders/npm-lock-v2/index.ts
+++ b/lib/dep-graph-builders/npm-lock-v2/index.ts
@@ -190,16 +190,23 @@ const getChildNode = (
   pkgKeysByName: Map<string, string[]>,
   overrides?: Overrides,
 ) => {
+  let version = depInfo.version;
+
   const override =
     overrides &&
-    checkOverrides(
-      [...ancestry, { name, version: depInfo.version }] as Ancestry[],
-      overrides,
-    );
+    checkOverrides([...ancestry, { name, version }] as Ancestry[], overrides);
+
+  if (override) {
+    version = override;
+  }
+
+  if (version.startsWith('npm:')) {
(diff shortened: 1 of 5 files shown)


lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -190,16 +190,23 @@ const getChildNode = (
...
-      overrides,
-    );
+    checkOverrides([...ancestry, { name, version }] as Ancestry[], overrides);
+
+  if (override) {
+    version = override;
+  }
+
+  if (version.startsWith('npm:')) {
+    version = version.split('@').pop() || version;
```
dotkas, line 204: Is `@` guaranteed to always exist in that string?

dotkas (approved):

lib/dep-graph-builders/npm-lock-v2/index.ts
dotkas: Ok, it returns undefined. Thought it would throw an error.

JamesPatrickGill: Yes, but if it doesn't you just take the full string anyway

JamesPatrickGill: Can change to a `lastIndexOf` check

dotkas: Nah, it's OK by me at least.

PR #223, opened 2024-05-09
feat: added support for pnpm lockfiles v9

gemaxim:
### What this does

Extended support for pnpm v9 lockfiles. Added tests to match.


lib/dep-graph-builders/pnpm/lockfile-parser/lockfile-parser.ts
```diff
@@ -55,28 +55,30 @@ export abstract class PnpmLockfileParser {
...
-          optionalDependencies: versionData.optionalDependencies,
+          dependencies: versionData.dependencies || {},
+          devDependencies: versionData.devDependencies || {},
+          optionalDependencies: versionData.optionalDependencies || {},
         };
         packages[`${pkg.name}@${pkg.version}`] = pkg;
       },
     );
-    return packages;
+    this.extractedPackages = packages;
```
dotkas, line 81: Why are we doing it like this? Functions with side-effects looks a bit bad to me. But maybe I don't understand it.

dotkas (approved):
LGTM except I don't get the impure function change.

lib/dep-graph-builders/pnpm/lockfile-parser/lockfile-parser.ts
gemaxim: becasue the new lockfile version specifies dependencies and other important data in the 'snapshots' field that's being populated in the subclass constructor, after calling the super constructor. But yes, it's better to return the value and assign it where I call the function.

gemaxim: fixed

PR #226, opened 2024-05-15
fix: replacing micromatch with minimatch due to CVEs

dotkas:
There's a bunfight going on over at Micromatch, to which I won't link, as it will show up in their conversation, but the link is https://github.com/micromatch/micromatch and the issue is `243` in case you're interested.

In any event that looks like it will never be sorted any time soon, so let's just fix the paperwork the easy way.

### DIFF
--- a/lib/dep-graph-builders/npm-lock-v2/index.ts
+++ b/lib/dep-graph-builders/npm-lock-v2/index.ts
@@ -4,8 +4,8 @@ import {
   PackageJsonBase,
   ProjectParseOptions,
 } from '../types';
-import { extractPkgsFromNpmLockV2 } from './extract-npm-lock-v2-pkgs';
 import type { NpmLockPkg } from './extract-npm-lock-v2-pkgs';
+import { extractPkgsFromNpmLockV2 } from './extract-npm-lock-v2-pkgs';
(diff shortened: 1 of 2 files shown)


gemaxim (commented):
I changed this import to pass linting.
Is there a reason you used require instead of import ... from?

lib/dep-graph-builders/npm-lock-v2/index.ts
```diff
@@ -18,9 +18,9 @@ import { OutOfSyncError } from '../../errors';
 import { LockfileType } from '../../parsers';
 
 import * as semver from 'semver';
-import * as micromatch from 'micromatch';
 import * as pathUtil from 'path';
 import { eventLoopSpinner } from 'event-loop-spinner';
+import * as multimatch from 'multimatch';
```
gemaxim, line 23: I changed this import to pass linting.
Is there a reason you used require instead of import ... from?

gemaxim (approved):

lib/dep-graph-builders/npm-lock-v2/index.ts
dotkas: It failed for me locally 🤷🏻

dotkas:
Looks like a fix was released, dropping this.

PR #235, opened 2024-07-24
fix: [OSM-1122] handle projects cycles in workspace

gemaxim:
### What this does

Problem 1 solved
Cyclic dependencies corresponding to projects inside a workspace were not handled correctly because resolving a local workspace project as a dependency of another enters a recursive normalisation function. 
Example of scneario
package.json - Project A
pnpm-lock.yaml
pnpm-workspace-yaml
packages/b/package.json - Project B
packages/c/package.json - Project C

,all being part of a workspace. A has B as a dependency. B has C as a dependency and C has B as a dependency.
See test workspace-cyclic.

Problem 2 solved
When building the dependency graph, if there is a cyclic dependency to the root package, this is now handled.
See test workspace-cyclic-root.

Some improvement: Projects of a workspace are not 'normalized' multiple times. For example, if a project B is a dependency of the root project, it is 'normalized' (dependencies are retrieved, along with name, version, etc.) once when the root is processed (as part of its dependencies) and stored. When B is handled as a project, it is not normalized anymore and just retrieved.

### Notes for the reviewer

See tests.

### DIFF
--- a/lib/dep-graph-builders/pnpm/lockfile-parser/lockfile-parser.ts
+++ b/lib/dep-graph-builders/pnpm/lockfile-parser/lockfile-parser.ts
@@ -10,6 +10,7 @@ import {
 } from '../types';
 import { valid } from 'semver';
 import * as pathUtil from 'path';
+import { isEmpty } from 'lodash';
 
 export abstract class PnpmLockfileParser {
   public lockFileVersion: string;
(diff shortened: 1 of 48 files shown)


lib/dep-graph-builders/pnpm/lockfile-parser/lockfile-parser.ts
```diff
@@ -96,9 +111,49 @@ export abstract class PnpmLockfileParser {
...
+        dependencies: this.topLevelDepsToNormalizedPkgs(prodDeps),
+        devDependencies: this.topLevelDepsToNormalizedPkgs(devDeps),
+        optionalDependencies: this.topLevelDepsToNormalizedPkgs(optionalDeps),
+        isDev: false,
+      };
+    }
     return { ...prodDeps, ...devDeps, ...optionalDeps, ...peerDeps };
   }
 
+  public normalizedPkgToTopLevel(pkg: NormalisedPnpmPkg): PnpmDeps {
```
andrei-cacio, line 128: nitpick-ish: names that begin past tense such as `normalizedBlaBla` are usually used for variables, constants, things that don't compute. For function is common to use present verbs like `normalizeBlaBla`

gemaxim: this function is basically just an object transformation from a format (the format for normalised packages) another format (PnpmDeps)

MarcusArdelean (approved):

PR #249, opened 2024-08-13
fix: fail early for top level deps out of sync pnpm and updated tests

gemaxim:
### What this does

We should throw OutOfSync early if top level deps were not found in extracted lockfile packages.

Updated tests for pnpm/npm/yarn: we shouldn't use try-catch to test a failing scenario because if if the function doesn't fail as expected the test still passes

### DIFF
--- a/lib/dep-graph-builders/pnpm/utils.ts
+++ b/lib/dep-graph-builders/pnpm/utils.ts
@@ -34,15 +34,6 @@ export const getPnpmChildNode = (
     resolvedVersion = pkgData.version;
   }
   if (!pkgs[childNodeKey]) {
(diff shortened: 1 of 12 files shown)


lib/dep-graph-builders/pnpm/utils.ts
```diff
@@ -34,15 +34,6 @@ export const getPnpmChildNode = (
     resolvedVersion = pkgData.version;
   }
   if (!pkgs[childNodeKey]) {
-    if (lockfileParser.isWorkspaceLockfile()) {
```
gemaxim, line 37: This is not needed anymore because workspace importers are added to extracted packages and we should fail with OutOfSync if not found there

dotkas (approved):

PR #265, opened 2025-03-19
feat: pnpm - function for parsing single project from a workspace

gemaxim:
Added function for parsing a single package.json within a workspace. Expected input : package.json file content, pnpm-lock.yaml file content, options, importer (the relative path to the package.json file)
Unfortunately, if there's no context/access to the other package.json files, we can't determine the versions of the referenced local projects - hence the added expected outputs in tests.

### DIFF
--- a/lib/dep-graph-builders/index.ts
+++ b/lib/dep-graph-builders/index.ts
@@ -13,7 +13,11 @@ import {
   extractPkgsFromYarnLockV2,
 } from './yarn-lock-v2';
 import { parseNpmLockV2Project } from './npm-lock-v2';
-import { parsePnpmProject, parsePnpmWorkspace } from './pnpm';
+import {
+  parsePnpmProject,
+  parsePnpmWorkspace,
+  parsePnpmWorkspaceProject,
+} from './pnpm';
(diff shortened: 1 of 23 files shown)


test/jest/dep-graph-builders/pnpm-lock.test.ts
```diff
@@ -10,23 +10,23 @@ describe.each(['pnpm-lock-v5', 'pnpm-lock-v6', 'pnpm-lock-v9'])(
...
-          'alias-sub-dependency',
-          'empty-project',
-          'git-protocol-peer-deps',
-          'codelab-ref-deps',
+          // 'cyclic-dep',
+          // // 'deeply-nested-packages',
+          // 'deeply-scoped',
+          // 'different-versions',
+          // 'local-pkg-without-workspaces',
+          // // 'dist-tag-sub-dependency',
```
andrei-cacio, line 20: can we just delete these?

gemaxim: ups, no
reverting now

andrei-cacio (approved):

PR #267, opened 2025-04-03
fix: handle out of sync in resolutions for yarn 2

gemaxim:
Handle OutOfSync for yarn resolutions - if the resolved package from resolution is not resolved in the lockfile, we need to handle as usual out of sync errors.


DOlufemi (approved):

lib/dep-graph-builders/yarn-lock-v2/utils.ts
```diff
@@ -118,6 +118,24 @@ export const getYarnLockV2ChildNode = (
 
   if (resolvedVersionFromResolution) {
     const childNodeKeyFromResolution = `${name}@${resolvedVersionFromResolution}`;
+    if (!pkgs[childNodeKeyFromResolution]) {
+      if (strictOutOfSync && !/^file:/.test(resolvedVersionFromResolution)) {
```
andrei-cacio, line 122: I don't have context, but at a first glance I wouldn't understand why we need the /^file:/ regex check. Maybe add the output in a const with an explicative name? 

If it's very straightforward once you have context then ignore this comment.

gemaxim: I'm not 100% sure,  only 99%, but if it has a version with 'file:./' it's a local package that we can't reach with an SCM scan and also don't wanna throw an out of sync error for. 
We use this in a lot of places, so I'll create another PR to add the /^file:/.test(resolvedVersionFromResolution) in a function with some explanation

PR #272, opened 2025-05-26
feat(OSM-2021): add support for multiple versions of the same dependency in Yarn resolutions

rdghe:
- [ ] Reviewed by Snyk team

### What this does
Add support for resolving multiple versions of the same dependency in Yarn resolutions.

### More information

- [OSM-2021](https://snyksec.atlassian.net/browse/OSM-2021)


lib/dep-graph-builders/yarn-lock-v2/utils.ts
```diff
@@ -104,16 +108,41 @@ export const getYarnLockV2ChildNode = (
...
+              }
+            }
+          }
+        } catch (e) {
+          debug(`Error parsing resolution key(${resKey}): ${e}$`);
+        }
+      }
+    }
+    // Check for global resolution by package name (e.g., "packageName": "version")
+    if (resolutions[name]) {
```
gemaxim, line 142: can we move this above the for loop for an early exit?

gemaxim: got it, thanks! so the more specific range match has prio

PR #300, opened 2026-03-06
fix: patch protocol URL decoding and npm alias package creation

james-snyk:
- [ ] Reviewed by Snyk team

## Problem

Users scanning Yarn v2/v3 workspaces with patch dependencies or npm aliases were experiencing two critical errors:

1. **"Dependency ... was not found in yarn.lock"** - When using patch protocol dependencies
2. **"Invalid depGraph" (SNYK-0003)** - When submitting dependency graphs with npm aliases to the Snyk API

### 1. URL Encoding Mismatch (Patch Protocol)
- package.json resolutions use URL-encoded strings: `patch:lodash@npm%3A4.17.21#...`
- yarn.lock keys are decoded: `patch:lodash@npm:4.17.21#...`
- Lookup was failing because we were using the encoded string to search decoded keys

### 2. npm Alias Package Identity Mismatch
- Alias packages were created with **target names** instead of **alias names**
- Node IDs used alias names: `string-width-cjs@4.2.3`
- Package names used target names: `string-width@4.2.3`
- This caused node-to-package reference mismatches, making the depGraph invalid

### Fix 1: Patch Protocol URL Decoding
**File:** lib/dep-graph-builders/yarn-lock-v2/utils.ts

- Decode URL-encoded resolution values before package lookup
- Use simple version format for node IDs (lodash@4.17.21) instead of full patch strings
- Prevents patch: protocol strings from appearing in node IDs

### Fix 2: npm Alias Package Creation
**File:** lib/dep-graph-builders/yarn-lock-v2/build-depgraph-simple.ts

- Use childNode.alias.aliasName when creating package nodes for aliases
- Ensures package IDs match node IDs (e.g., both use string-width-cjs@4.2.3)
- Creates separate package entries for aliases with correct identity


lib/dep-graph-builders/yarn-lock-v2/utils.ts
```diff
@@ -174,38 +174,47 @@ export const getYarnLockV2ChildNode = (
...
-      if (strictOutOfSync && !/^file:/.test(resolvedVersionFromResolution)) {
+      if (strictOutOfSync && !/^file:/.test(decodedResolution)) {
         throw new OutOfSyncError(
           childNodeKeyFromResolution,
           LockfileType.yarn2,
         );
       } else {
         return {
-          id: childNodeKeyFromResolution,
+          id: `${name}@${decodedResolution}`,
```
cjheppell: L180 already constructs this, so I think we can use `childNodeKeyFromResolution` instead of the string interpolation.

lib/dep-graph-builders/yarn-lock-v2/utils.ts
```diff
@@ -174,38 +174,47 @@ export const getYarnLockV2ChildNode = (
...
         };
       }
     }
 
+    const pkgData = pkgs[childNodeKeyFromResolution];
+    if (!pkgData) {
+      debug(`Package not found for key: ${childNodeKeyFromResolution}`);
+      throw new OutOfSyncError(childNodeKeyFromResolution, LockfileType.yarn2);
+    }
+
```
cjheppell, line 208: We weren't checking this before - was that an oversight?
Is there a possibility that this could cause behaviour changes if before we tolerated a missing package? (I suspect _not tolerating_ a missing package is better though)

test/jest/dep-graph-builders/yarn-lock-v2.test.ts
```diff
@@ -632,10 +632,109 @@ describe('yarn.lock v2 parsing', () => {
       },
       {
         nodeId: 'lib-types@1.0.0',
-        pkgId: '@scope/types@1.0.0',
+        pkgId: 'lib-types@1.0.0',
```
cjheppell, line 635: why the change from `@scope/` to `lib-`?

cjheppell (commented):
Some comments/questions.  Not blockers - just trying to understand rationale behind some changes and also if the newly thrown OutOfSyncError is something that we were missing before.

test/jest/dep-graph-builders/yarn-lock-v2.test.ts
james-snyk: Due to the alias change before we set the alias as the nodeId and the target as the pkgId but this caused packages not get to linked up correctly and caused errors. Alias information is preservered in the alais field for vuln look ups

lib/dep-graph-builders/yarn-lock-v2/utils.ts
james-snyk: This was me not double checking but we do a check above what basically makes this redundant.

cjheppell (approved):

PR #305, opened 2026-04-30
fix: use actual package name from lockfile for aliased packages

james-snyk:
- [ ] Reviewed by Snyk team

## Problem

When a transitive dependency resolves to an aliased package in node_modules,
the parser was incorrectly using the folder name (alias) instead of the actual
package name from the lockfile.

Example scenario:
- `plotly.js` declares: `"regl": "npm:@plotly/regl@^2.1.2"` (explicit alias)
- `gl-text` declares: `"regl": "^2.0.0"` (regular dependency, not an alias)
- Both resolve to `node_modules/regl` which contains `@plotly/regl`

The parser was returning `regl@2.1.2` instead of `@plotly/regl@2.1.2` for
the `gl-text` dependency, causing:
- Incorrect vulnerability scanning (checking wrong package)
- Incorrect SBOM generation (reporting wrong package)
- Incorrect license compliance (checking wrong package)

## Solution

Use the `name` field from the lockfile entry as the authoritative source for
the actual package name. According to npm lockfile v2/v3 specification, the
`name` field contains the actual package name, which may differ from the
installation path for aliased packages.

### DIFF
--- a/lib/dep-graph-builders/npm-lock-v2/index.ts
+++ b/lib/dep-graph-builders/npm-lock-v2/index.ts
@@ -96,16 +96,25 @@ export const buildDepGraphNpmLockV2 = async (
 
   const pkgKeysByName: Map<string, string[]> = Object.keys(npmLockPkgs).reduce(
     (acc, key) => {
-      const name = key.replace(/.*node_modules\//, '');
-      if (!name) {
+      const folderName = key.replace(/.*node_modules\//, '');
+      if (!folderName) {
         return acc;
       }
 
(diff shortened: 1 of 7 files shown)


JamesPatrickGill (commented):
2 general questions before deeper review:

1. Is `name` always present? I'm not sure it's a required field?
2. This changes a fair whack of expected.json files. Do we need to be concerned of how many projects will now return different results for rollout? Whats the plan?

JamesPatrickGill (approved):
Thanks for tightening up

